PR #1, opened 2021-02-09
regenrate controller code after repository split

kumargauravsharma:
Issue #, if available:

The ElastiCache service controller code needs to be regenerated per the ACK split repositories and updated import paths.
Check compilation and unit tests by running `make test` for `elasticache-controller`.

Description of changes:

* Updated `elasticache-controller/go.mod` file to refer to local `runtime`
* Regenerated ElastiCache controller using 
```
cd $GOPATH/src/github.com/aws-controllers-k8s/code-generator
make build-ack-generate
export SERVICE="elasticache"
make build-controller SERVICE=$SERVICE
```

Testing: `make test` passed.


go.mod
```diff
@@ -2,8 +2,10 @@ module github.com/aws-controllers-k8s/elasticache-controller
 
 go 1.14
 
+replace github.com/aws-controllers-k8s/runtime => ../runtime
```
A-Hilaly, line 5: I don't think we should keep the `replace` statement here, it should be used only for dev purposes. If the `runtime` repository doesn't exist locally, `go build` will complain with `...runtime/go.mod: no such file or directory`. Other than that there is also the risk of using an outdated version/commit of `runtime`.

A-Hilaly (approved):
LGTM overall 👍. I left one comment

go.mod
jaypipes: @A-Hilaly is correct on the above.

kumargauravsharma: Updated the go.mod file.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/aws-controllers-k8s/elasticache-controller
 go 1.14
 
 require (
-	github.com/aws/aws-controllers-k8s v0.0.2
+	github.com/aws-controllers-k8s/runtime v0.0.0-20210204203051-91a6bd53a48c
```
jaypipes, line 6: :+1:

jaypipes (approved):
:+1:

PR #6, opened 2021-02-17
Enable mocking ElastiCache API in unit tests

kumargauravsharma:
Issue #, if available:

To unit test custom code there is a need to mock ElastiCache service API.

Description of changes:

* Adds `MockElastiCacheAPI` and reference implementation to mock `DecreaseReplicaCountWithContext`, `IncreaseReplicaCountWithContext` in it. Additional APIs can be mocked as needed.
* Adds a unit test to validate the API mock.

This will help write unit tests for the custom code which was otherwise not possible (without mocking service API) due to failing service api calls during unit tests.

`make test` passed.


mockapi/mock_elasticache_api.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (m MockElastiCacheAPI) IncreaseReplicaCountWithContext(
+	aws.Context,
+	*svcsdk.IncreaseReplicaCountInput,
+	...request.Option,
+) (*svcsdk.IncreaseReplicaCountOutput, error) {
+	return m.MockIncreaseReplicaCountOutput, m.MockIncreaseReplicaCountError
+}
+
+// add more API methods implementation to MockElastiCacheAPI as needed for unit testing
```
jaypipes, line 53: Instead of doing this, how about use the `mockery` tool to automatically generate a mocked ElasticacheAPI struct?

jaypipes (changes requested):
I think it's best to use `mockery` to automatically generate a mock of the SDK API struct for ElastiCache. That way, you won't need to worry about manually updating the mock when the upstream aws-sdk-go interface changes, and you can use the excellent "github.com/stretchr/testify/mock" library and all of its assertion functions.

Grab the mockery installation and auto-generation targets from the runtime Makefile here:

https://github.com/aws-controllers-k8s/runtime/blob/bfefd1d1001dcb3af7c311d9837cea6aee637be3/Makefile#L16-L22

and adapt that to point mockery at, say, $GOPATH/pkg/mod/github.com/aws/aws-sdk-go@v1.35.5/service/elasticache/elasticacheiface and output the full mocked API.

Something akin to this:

```bash
pushd $GOPATH/pkg/mod/github.com/aws/aws-sdk-go@v1.35.5/service/elasticache/elasticacheiface 1>/dev/null

$MOCKER_BIN_PATH --all --dir=. --output=$SERVICE_CONTROLLER_SOURCE_PATH/mocks/aws-sdk-go/elasticache/

popd 1>/dev/null
```

mockapi/mock_elasticache_api.go
kumargauravsharma: Thanks for your inputs @jaypipes 
I have updated the code to use `mockery`.

Makefile
```diff
@@ -3,6 +3,9 @@ SHELL := /bin/bash # Use bash syntax
 # Set up variables
 GO111MODULE=on
 
+ELASTICACHE_API_PATH="$(shell echo $(GOPATH))/pkg/mod/github.com/aws/aws-sdk-go@v1.37.4/service/elasticache/elasticacheiface"
```
jaypipes, line 6: approving this for now, but you'll likely want to do a little `awk` or `grep`-fu to replace the `1.37.4` hard-coding above by looking in the go.mod file :)

jaypipes (approved):
awesome, thanks Kumar!

jaypipes:
@kumargauravsharma the unit-test workflow is failing due to this:

```
Installing mockery into bin/mockery ... ok.
/bin/bash: line 0: pushd: /pkg/mod/github.com/aws/aws-sdk-go@v1.37.4/service/elasticache/elasticacheiface: No such file or directory
```

Looks like $GOPATH isn't defined in the Github Actions default runner -- either that or we're going to need to figure out some other way of specifying the path to the Go module source for the aws-sdk-go package...

kumargauravsharma:
> @kumargauravsharma the unit-test workflow is failing due to this:
> 
> ```
> Installing mockery into bin/mockery ... ok.
> /bin/bash: line 0: pushd: /pkg/mod/github.com/aws/aws-sdk-go@v1.37.4/service/elasticache/elasticacheiface: No such file or directory
> ```
> 
> Looks like $GOPATH isn't defined in the Github Actions default runner -- either that or we're going to need to figure out some other way of specifying the path to the Go module source for the aws-sdk-go package...

Fixed the issue and and GitHub action completed with success.

jaypipes (approved):

PR #10, opened 2021-02-23
mock api output from json file for unit tests

kumargauravsharma:
Issue #, if available:

Description of changes:

This PR adds methods to create mock ElastiCache API output from json file.
This would help unit test scenarios by adding corresponding API response json from file.

### DIFF
--- a/pkg/resource/replication_group/custom_update_api_test.go
+++ b/pkg/resource/replication_group/custom_update_api_test.go
@@ -16,11 +16,13 @@ package replication_group
 import (
 	"context"
 	"fmt"
+	"github.com/aws-controllers-k8s/elasticache-controller/pkg/testutil"
 	ackmetrics "github.com/aws-controllers-k8s/runtime/pkg/metrics"
 	"github.com/aws-controllers-k8s/runtime/pkg/requeue"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/mock"
 	"go.uber.org/zap/zapcore"
+	"path/filepath"
 	ctrlrtzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


mocks/aws-sdk-go/elasticache/ElastiCacheAPIOutput.go
```diff
@@ -0,0 +1,79 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
jaypipes, line 1: A couple things about this that I would suggest changing.

You definitely don't want to place any of this code into the `mocks/` directory. Why? Because of this:

https://github.com/aws-controllers-k8s/elasticache-controller/blob/1313e661427da6c2a886d82d72d7fedb3187dc4a/Makefile#L29-L30

:)

First, I would suggest putting generic test utility code into `pkg/testutil` package. I recommend placing all test fixtures (the .json files) into a `pkg/resource/$RESOURCE/testdata` directory. See [this post by Dave Cheney](https://dave.cheney.net/2016/05/10/test-fixtures-in-go) about why you should call the directory `testdata`.

Secondly, you may not need functions like the `newMockOutput` or `getFileContents` functions if you created a generic fixture loader in `pkg/testutil` that looks something like this:

```go
package testutil

import (
	"encoding/json"
	"io/ioutil"
)

// LoadFromFixture fills an empty pointer variable with the
// data from a fixture JSON file.
func LoadFromFixture(
	fixturePath string,
	output interface{}, // output should be an addressable type (i.e. a pointer)
) {
	contents, err := ioutil.ReadFile(fixturePath)
	if err != nil {
		panic(err)
	}
	err = json.Unmarshal(contents, output)
	if err != nil {
		panic(err)
	}
}
```

You could then call this function from `pkg/resource/replicationgroup/custom_update_api_test.go`:

```go
...
	import "github.com/aws-controllers-k8s/elasticache-controller/pkg/testutil"
...
	var mockOutput svcsdk.DecreaseReplicaCountOutput
	testutil.LoadFromFixture("testdata/DecreaseReplicaCountOutput.json", &mockOutput)
```

jaypipes (changes requested):
Kumar, please see inline comment for suggestions.

mocks/aws-sdk-go/elasticache/ElastiCacheAPIOutput.go
kumargauravsharma: Thanks for your suggestions Jay. The resultant code is simpler.
I have updated the changes in this PR.

jaypipes (approved):
cool. :)

PR #12, opened 2021-02-25
remove global replication group id from replication group input

kumargauravsharma:
Issue #, if available:

`GlobalReplicationGroup` is ignored resource in `generator.yaml`, it is not yet supported custom resource.
Thus, `GlobalReplicationGroupId` needs to be removed from `ReplicationGroup` input `Spec` (and `CreateReplicationGroupInput`)

Description of changes:

* Updated `generator.yaml` by listing `CreateReplicationGroupInput.GlobalReplicationGroupId` as `ignore.field_paths`
* Regenerated the service controller code
* `make test` passed.


jaypipes (approved):
k, makes sense. aren't you glad you're in the "0" series still? :)

PR #18, opened 2021-04-03
change test framework behavior to allow more control over assertions, increase replication group unit test coverage

echen-98:
Description of changes:

### Changes to the test framework assertion behavior
For expectations in `test_suite.yaml`, `latest_state` and `error` are asserted independently. The test framework now ensures that at least one explicit assertion exists (there can't be no assertions for either `latest_state` or `error`). 

In any other case:
- If the key (`latest_state` or `error`) is left out completely or has an empty value in `test_suite.yaml`, this will be interpreted as "no explicit expectation", and no assertions will run.
- If the key has value "nil", the test framework will assert that the relevant object is `nil`
- Otherwise, the framework will assert that the actual result of the test aligns with the expected result specified in the value of `latest_state` and/or `error`.

This means that `setupExpectationContext` has been removed and the test suite runner more or less works directly from the test suite config in `test_suite.yaml` so that nothing gets lost in translation.

Further, I added the ability to mock aws-sdk-go error types so that the correct code paths get executed for negative cases.

### Other test cases
Added some new test cases covering Create, ReadOne, and Delete scenarios, including some negative tests.

### Changes I will follow up with soon
Additional negative cases and cases covering the Update operation.


pkg/testutil/test_suite_runner.go
```diff
@@ -96,31 +95,55 @@ func (runner *TestSuiteRunner) runTestScenario(scenarioName string, fixtureCxt *
...
+ */
+func (runner *TestSuiteRunner) assertExpectations(assert *assert.Assertions, expectation *Expect, actual acktypes.AWSResource, err error) {
+	if expectation.LatestState == "" && expectation.Error == "" {
+		fmt.Println("Invalid test case: no expectation given for either latest_state or error")
+		assert.True(false)
+		return
+	}
+
+	// expectation exists for at least one of LatestState and Error; assert results independently
+	if expectation.LatestState == "nil" {
```
jaypipes, line 121: Instead of checking for a string value called "nil", you could use the `omitempty` Go tag on the [`Expect.LatestState` and `Expect.Error` fields](https://github.com/aws-controllers-k8s/elasticache-controller/blob/f07e1b2a5184c65a1c0039fa20778f6436489757/pkg/testutil/test_suite_config.go#L60-L61) after making those fields' Go types a string pointer:

```go
// Expect contains the expected resource state or error after
// executing the operation under test. Either LatestState or Error
// should be non-nil.
type Expect struct {
	LatestState *string `json:"latest_state,omitempty"`
	Error       *string `json:"error,omitempty"`
}
```

And then here in the `assertExpectations()` you can simply check to see `if expectation.LatestStatus == nil`. That way, your `test_suite.yaml` is shorter and easier to read, since you only need to set either a `latest_state` or an `error` field, depending on whether the expected outcome is an error or not.

pkg/testutil/test_suite_runner.go
```diff
@@ -96,31 +95,55 @@ func (runner *TestSuiteRunner) runTestScenario(scenarioName string, fixtureCxt *
...
 		}
-		// Delta only contains `Spec` differences. Thus, need to have Delegate.Equal to compare `Status`.
-		assert.True(runner.Delegate.Equal(expectationCxt.latest, actual))
+
+		// Delta only contains `Spec` differences. Thus, we need Delegate.Equal to compare `Status`.
+		assert.True(runner.Delegate.Equal(expectedLatest, actual))
+	}
+
+	if expectation.Error == "nil" {
+		assert.Nil(err)
```
jaypipes, line 141: And here, you can just do:

```go
if expectation.Error == nil {
    assert.Nil(err)
} else {
    expectedError := errors.New(expectation.Error)
    assert.NotNil(err)
    assert.Equal(expectedError.Error(), err.Error())
}
```

jaypipes (changes requested):
@echen-98 I'm loving these changes, awesome work! I have a suggestion inline for how to make the YAML files a bit easier to read/be shorter, though. Please let me know what you think.

pkg/testutil/test_suite_runner.go
echen-98: Thanks for the suggestion re: the `omitempty` tag – my one concern is that, in addition to asserting LatestState and Error independently, we need to support the following behaviors:

1. Assert that LatestState or Error matches the given value
2. Assert that LatestState or Error are nil
3. Don't care about the value of LatestState or Error

The third case is the problematic one. There are some scenarios where the call to a resource manager's CRUD operation results in an error, but the returned resource is actually non-`nil` (e.g. the first test case `ReadOne=DNE` in `test_suite.yaml`). I think in this case we only care about asserting the error, and that we don't necessarily care about the `nil`-ness of the resource. 

Our initial implementation failed here because it assumed that every time the error was non-nil, the returned resource had to be `nil` (i.e both these assertions were coupled together in the same place). 

The block of code you suggested the below for asserting the `expectation.Error` should work with the `omitempty` approach, since we pretty much always care about the value of the Error, but how can we specify that we don't care about the value of `expectation.LatestState`? Changing the expectation struct to pointers to strings would allow us to do the following:

1. Omit the key entirely in test config -> `Expect.LatestState` is `nil`
2. Include the key with an empty value in test config -> `Expect.LatestState` is non-`nil`, `*Expect.LatestState` is `""`
3. Include the key with a nonempty value in test config -> `Expect.LatestState` is non-`nil`, *Expect.LatestState` is non-empty

We could map (1) to "explicitly assert nil" and (2) to "don't care about the value" to get rid of all the extra `nil` assertion lines. However, this would make the test config very confusing to read, and I really don't think we should sacrifice clarity for brevity in this case. 

Alternatively, we could use the more intuitive mapping of (1) to "don't care about the value" and (2) to "explicitly assert nil", but this would be just as long as the current implementation, and more confusing.

So in summary I think we should keep it as-is since it's the clearest/most intuitive way I can think of to encode the expectations of the test writer, and that the marginal gain of each scenario being 1-2 lines shorter in the test config is not worth the added confusion. Please let me know what you think or if I misinterpreted the suggestion!

jaypipes (approved):
@echen-98 I do see your point, yes. I will noodle on this a bit and see if I can come up with a little bit more elegant approach, but I won't hold this up. @kumargauravsharma please feel free to approve and merge.

echen-98:
@jaypipes thanks, if we figure out something better later then I'm all for it

kumargauravsharma (approved):

kumargauravsharma:
@echen-98 unit test have failed for this PR. I request you to fix it to enable merging.

echen-98:
@kumargauravsharma fixed the failing unit test in `manager_test.go`. I also added the error mocking implementation to unblock the unfailing negative declarative tests, please review (especially `util.go`, `setupFixtureContext` in `test_suite_runner.go`, and `test_suite_config.go`)

kumargauravsharma (approved):

kumargauravsharma:
Thanks @echen-98

PR #20, opened 2021-04-14
add replication group declarative tests, remove unused argument from engineVersionsDiffer

echen-98:
Description of changes:

- increase replication group code coverage (probably best to focus review on `test_suite.yaml` since there are many `testdata` files)
- small tweak to engineVersionsDiffer: removed the second argument as it was unused

### DIFF
--- a/pkg/resource/replication_group/custom_update_api.go
+++ b/pkg/resource/replication_group/custom_update_api.go
@@ -134,7 +134,7 @@ func (rm *resourceManager) modifyReplicationGroup(
 
 	// SecurityGroupIds, EngineVersion
 	if rm.securityGroupIdsDiffer(desired, latest, latestCacheCluster) ||
-		rm.engineVersionDiffer(desired, latest, latestCacheCluster) {
+		rm.engineVersionsDiffer(desired, latestCacheCluster) {
 		input := rm.newModifyReplicationGroupRequestPayload(desired, latest, latestCacheCluster)
 		resp, respErr := rm.sdkapi.ModifyReplicationGroupWithContext(ctx, input)
 		rm.metrics.RecordAPICall("UPDATE", "ModifyReplicationGroup", respErr)
@@ -623,7 +623,7 @@ func (rm *resourceManager) describeCacheCluster(
 		}
 		return cc, nil
 	}
-	return nil, nil
+	return nil, fmt.Errorf("could not find a non-nil cache cluster from API response")
 }
 
 // securityGroupIdsDiffer return true if
@@ -701,7 +701,7 @@ func (rm *resourceManager) newModifyReplicationGroupRequestPayload(
 		input.SetSecurityGroupIds(ids)
 	}
 
-	if rm.engineVersionDiffer(desired, latest, latestCacheCluster) &&
+	if rm.engineVersionsDiffer(desired, latestCacheCluster) &&
 		desired.ko.Spec.EngineVersion != nil {
 		input.SetEngineVersion(*desired.ko.Spec.EngineVersion)
 	}
@@ -709,11 +709,16 @@ func (rm *resourceManager) newModifyReplicationGroupRequestPayload(
 	return input
 }
 
-// engineVersionDiffer return true if
-// Engine Version differs between desired spec and latest (from cache cluster) status
-func (rm *resourceManager) engineVersionDiffer(
+/*
+engineVersionsDiffer returns true if the desired engine version is different
+from the latest observed engine version. Inputs:
+
+desired: the resource representing the desired state
+latestCacheCluster: a CacheCluster object representing one cache node in the replication group, which
+	reveals the currently used engine version for the entire replication group
+ */
+func (rm *resourceManager) engineVersionsDiffer(
 	desired *resource,
-	latest *resource,
 	latestCacheCluster *svcsdk.CacheCluster,
 ) bool {
 	if desired.ko.Spec.EngineVersion == nil {
--- a/pkg/resource/replication_group/custom_update_api_test.go
+++ b/pkg/resource/replication_group/custom_update_api_test.go
@@ -966,9 +966,9 @@ func provideCacheClusterSecurityGroups(IDs ...string) []*svcsdk.SecurityGroupMem
 	return securityGroups
 }
 
-// TestEngineVersionDiffer tests scenarios to check if desired, latest (from cache cluster)
+// TestEngineVersionsDiffer tests scenarios to check if desired, latest (from cache cluster)
 // Engine Version configuration differs.
-func TestEngineVersionDiffer(t *testing.T) {
+func TestEngineVersionsDiffer(t *testing.T) {
 	assert := assert.New(t)
 	require := require.New(t)
 	// setup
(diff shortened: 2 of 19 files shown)


pkg/resource/replication_group/custom_update_api.go
```diff
@@ -701,19 +701,24 @@ func (rm *resourceManager) newModifyReplicationGroupRequestPayload(
...
-func (rm *resourceManager) engineVersionDiffer(
+/*
+engineVersionsDiffer returns true if the desired engine version is different
+from the latest observed engine version. Inputs:
+
+desired: the resource representing the desired state
+latestCacheCluster: a CacheCluster object representing one cache node in the replication group, which
+	reveals the currently used engine version for the entire replication group
+ */
+func (rm *resourceManager) engineVersionsDiffer(
```
jaypipes, line 720: Note that if ELC was using the Delta struct that is passed into the sdkUpdate code paths, this would be as simple as:

```go
if delta.DifferentAt("Spec.EngineVersion")
```

pkg/resource/replication_group/custom_update_api_test.go
```diff
@@ -966,59 +966,55 @@ func provideCacheClusterSecurityGroups(IDs ...string) []*svcsdk.SecurityGroupMem
 	return securityGroups
 }
 
-// TestEngineVersionDiffer tests scenarios to check if desired, latest (from cache cluster)
+// TestEngineVersionsDiffer tests scenarios to check if desired, latest (from cache cluster)
 // Engine Version configuration differs.
-func TestEngineVersionDiffer(t *testing.T) {
+func TestEngineVersionsDiffer(t *testing.T) {
```
jaypipes, line 971: And this test could be deleted.

pkg/resource/replication_group/testdata/replication_group/cr/rg_cmd_before_engine_version_upgrade.yaml
```diff
@@ -0,0 +1,94 @@
...
+      message: Replication group rg-cmd created
+  globalReplicationGroupInfo: {}
+  memberClusters:
+    - rg-cmd-001
+    - rg-cmd-002
+  multiAZ: disabled
+  nodeGroups:
+    - nodeGroupID: "0001"
+      nodeGroupMembers:
+        - cacheClusterID: rg-cmd-001
```
jaypipes, line 72: Just curious, but if cache mode is disabled, how are the RG node groups related to a cache cluster?

jaypipes (approved):
A couple comments inline but nothing major.

pkg/resource/replication_group/custom_update_api.go
echen-98: @jaypipes I'm glad you pointed this out since this gets to an issue that I'm looking into this week, which regards our ReplicationGroup manager/custom code but affects all our other CRs as well.

After the Delta changes to the runtime, if a user specifies a new `Spec.EngineVersion`, `desired.Spec` and `latest.Spec` will be identical and the resource manager's Update will not be invoked. 

IIRC (before the Delta changes), Update was invoked more often so our custom code has logic to determine whether the current state of the resource is up to date with the desired state, and whether to actually make the appropriate Modify API call. 

For example, in this case, our custom Update code identifies one of the cache nodes in the replication group, makes a [DescribeCacheClusters](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/elasticache/describe-cache-clusters.html) API call to get the properties of that cache node, and extracts the engine version from the response (since this information is not directly available from DescribeReplicationGroups, which is called in ReadOne). It then compares this to the desired `Spec.EngineVersion` and then chooses whether or not to invoke an appropriate ModifyReplicationGroup call.

So the question is, do you think this kind of logic should be moved to within the manager's ReadOne? i.e. it would be ReadOne's job to figure out the current engine version and update `latest.Spec.EngineVersion` to reflect this. If a new config was applied, `latest.Spec` and `desired.Spec` would no longer be identical so Update would be invoked and make a Modify API call according to the `Spec` differences.

In other words, should it be ReadOne's job to update every `Spec` and `Status` field possible (instead of letting Update implicitly figure out if there is a mismatch between the desired and latest observed states)?

pkg/resource/replication_group/testdata/replication_group/cr/rg_cmd_before_engine_version_upgrade.yaml
echen-98: @jaypipes Sorry I'm not quite following, did you mean cluster mode? 
"Cluster mode disabled (CMD)" = 1 node group
"Cluster mode enabled (CME)" = 2-500 node groups (but can be 1 node group as long as it's created with a `.cluster.on` parameter group...)

Where each node group has 1 primary node and 0-5 replica nodes, and nodes are synonymous with "cache clusters", at least for Redis (which is confusing because the entire replication group or one of its node groups can be considered a "cache cluster" in some sense).  @nmvk please correct me if I made any errors

I realize this is just clarifying terminology and may not answer your question, but wanted to make sure we're on the same page since I didn't quite understand the question.

pkg/resource/replication_group/custom_update_api.go
jaypipes: > @jaypipes I'm glad you pointed this out since this gets to an issue that I'm looking into this week, which regards our ReplicationGroup manager/custom code but affects all our other CRs as well.
> 
> After the Delta changes to the runtime, if a user specifies a new `Spec.EngineVersion`, `desired.Spec` and `latest.Spec` will be identical and the resource manager's Update will not be invoked.

Not sure why you think this is the case... the `desired.Spec` will contain the updated EngineVersion and the `latest.Spec` will contain the existing EngineVersion.

> IIRC (before the Delta changes), Update was invoked more often so our custom code has logic to determine whether the current state of the resource is up to date with the desired state, and whether to actually make the appropriate Modify API call.

The change you are referring to is that before the Delta changes, the `AWSResourceManager.Update` was called even when there was a change to *any* field on the desired or latest CR, *even* changes to `Status` fields. *After* the Delta changes, `AWSResourceManager.Update` is called *only* when there is a change in the `Spec` fields.

> For example, in this case, our custom Update code identifies one of the cache nodes in the replication group, makes a [DescribeCacheClusters](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/elasticache/describe-cache-clusters.html) API call to get the properties of that cache node, and extracts the engine version from the response (since this information is not directly available from DescribeReplicationGroups, which is called in ReadOne). It then compares this to the desired `Spec.EngineVersion` and then chooses whether or not to invoke an appropriate ModifyReplicationGroup call.

This is an unfortunate API. :(

It sounds like we will need to have a generic hook code for `sdk_read_one_post_set_output` that sets the value of `ko.Spec.EngineVersion` to this value determined by inspecting CacheNode objects. That way, the check for `desired.Spec == latest.Spec ` will fail in the reconciler and the `AWSResourceManager.Update` method will be called.

> So the question is, do you think this kind of logic should be moved to within the manager's ReadOne? i.e. it would be ReadOne's job to figure out the current engine version and update `latest.Spec.EngineVersion` to reflect this. If a new config was applied, `latest.Spec` and `desired.Spec` would no longer be identical so Update would be invoked and make a Modify API call according to the `Spec` differences.

Yes. The `sdk_read_one_post_set_output` hook is where you'd want to put this code.

> In other words, should it be ReadOne's job to update every `Spec` and `Status` field possible (instead of letting Update implicitly figure out if there is a mismatch between the desired and latest observed states)?

Yes, ReadOne should return the latest observed representation of the resource, and if the way to determine the latest "EngineVersion" (which is a Spec field because it's settable in Create) is to make some additional calls to DescribeCacheCluster, then that's what the `sdk_read_one_post_set_output` hook point is good for.

pkg/resource/replication_group/testdata/replication_group/cr/rg_cmd_before_engine_version_upgrade.yaml
jaypipes: @echen-98 it's just super confusing ElastiCache API, frankly...

If something is "cluster mode disabled", it is a surprising API user experience to see something called "cacheClusterID" in that something's attributes.

pkg/resource/replication_group/custom_update_api.go
jaypipes: I will just point out why I say this is an unfortunate API. If the ElastiCache DescribeReplicationGroup API call returned an EngineVersion field in its response, none of this would be necessary at all.

echen-98: @jaypipes thanks for clarifying the way forward, this is very helpful. 

Just wanted to clear up one point though:

> After the Delta changes to the runtime, if a user specifies a new Spec.EngineVersion, desired.Spec and latest.Spec will be identical and the resource manager's Update will not be invoked.

> > Not sure why you think this is the case... the desired.Spec will contain the updated EngineVersion and the latest.Spec will contain the existing EngineVersion.

I agree this is the desired behavior; all I'm saying is that with the controller in its current state, both `desired.Spec` and `latest.Spec` will have the updated EngineVersion so there will be no spec difference and therefore `Update` will not be invoked (I verified this via the debugger).  After `desired` is [copied into `ko`](https://github.com/aws-controllers-k8s/elasticache-controller/blob/1581529d917750ef9d9d18344b6b9bd14e68b074/pkg/resource/replication_group/sdk.go#L65), `ko.EngineVersion` is [never set to the current EngineVersion](https://github.com/aws-controllers-k8s/elasticache-controller/blob/1581529d917750ef9d9d18344b6b9bd14e68b074/pkg/resource/replication_group/sdk.go#L68-L286) because EngineVersion is not returned in the DescribeReplicationGroup API response as we discussed above. 

We'll update ReadOne with the hook as outlined above to resolve this.

echen-98: > I will just point out why I say this is an unfortunate API. If the ElastiCache DescribeReplicationGroup API call returned an EngineVersion field in its response, none of this would be necessary at all.

I think "unfortunate" may be a little generous... to give another example, `replicasPerNodeGroup` is another `spec`/`CreateReplicationGroup` field not directly returned by DescribeReplicationGroups, so we have custom code to determine the number of replicas by manually counting them from the `NodeGroups` object in the response: https://github.com/aws-controllers-k8s/elasticache-controller/blob/1581529d917750ef9d9d18344b6b9bd14e68b074/pkg/resource/replication_group/custom_update_api.go#L153-L258

We will probably deal with this in the same way.

jaypipes: > We will probably deal with this in the same way.

:+1:  :)

pkg/resource/replication_group/testdata/replication_group/cr/rg_cmd_before_engine_version_upgrade.yaml
echen-98: I think part of the confusion with the term "cache cluster" is that with Redis, it's just a single cache node, but with Memcached, it can be 1-20 nodes. But yeah things could probably be named more clearly.

PR #22, opened 2021-04-24
modify customSetOutput to update latest.Spec fields where possible

echen-98:
Description of changes:
In many cases, `latest.Spec` doesn't reflect the latest observed state for fields that the Elasticache API does not directly return. For example, we have to count the replicas from the `NodeGroups` portion of the API response to update `latest.Spec.ReplicasPerNodeGroup`, and we have to make an `DescribeCacheClusters` API call to retrieve information like the engine version (another field in `latest.Spec`.

The point of this PR is to initiate the process of updating all `latest.Spec` fields via `customSetOutput` so that `latest.Spec` actually reflects the latest observed state of the resource, therefore allowing the resource manager's `Update` to be invoked when appropriate. This logic was previously (and for the most part, still is) handled within the replication group resource manager's `Update`.

Spec fields covered:
- `ReplicasPerNodeGroup`
- `EngineVersion`

Changes were tested via the `make test-cover` target, and via a new e2e test in `test_replicationgroup.py`.

### DIFF
--- a/pkg/resource/replication_group/custom_set_output.go
+++ b/pkg/resource/replication_group/custom_set_output.go
@@ -143,6 +143,30 @@ func (rm *resourceManager) customSetOutput(
 		ko.Status.AllowedScaleDownModifications = nil
 		ko.Status.AllowedScaleUpModifications = nil
 	}
+
+	// populate relevant ko.Spec fields with observed state of respRG.NodeGroups
+	if respRG.NodeGroups != nil {
+		// if this is specified, all node groups should have the same # replicas so use the first
+		if ko.Spec.ReplicasPerNodeGroup != nil {
+			nodeGroup := respRG.NodeGroups[0]
+			if nodeGroup != nil && nodeGroup.NodeGroupMembers != nil {
+				if len(nodeGroup.NodeGroupMembers) > 0 {
+					*ko.Spec.ReplicasPerNodeGroup = int64(len(nodeGroup.NodeGroupMembers) - 1)
+				}
+			}
+		}
+
+		//TODO: set node group configuration in else if block
+	}
+
+	// updating some Spec fields requires a DescribeCacheClusters call
+	latestCacheCluster, err := rm.describeCacheCluster(ko)
+	if err == nil && latestCacheCluster != nil {
+		// only update EngineVersion if it's specified, and if latest state has a non-nil EngineVersion
+		if ko.Spec.EngineVersion != nil && latestCacheCluster.EngineVersion != nil {
+			*ko.Spec.EngineVersion = *latestCacheCluster.EngineVersion
+		}
+	}
 }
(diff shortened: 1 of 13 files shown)


pkg/resource/replication_group/custom_set_output.go
```diff
@@ -143,6 +143,30 @@ func (rm *resourceManager) customSetOutput(
...
+			nodeGroup := respRG.NodeGroups[0]
+			if nodeGroup != nil && nodeGroup.NodeGroupMembers != nil {
+				if len(nodeGroup.NodeGroupMembers) > 0 {
+					*ko.Spec.ReplicasPerNodeGroup = int64(len(nodeGroup.NodeGroupMembers) - 1)
+				}
+			}
+		}
+
+		//TODO: set node group configuration in else if block
+	}
```
jaypipes, line 160: Why are we overriding the value of `Spec.ReplicasPerNodeGroup` with the latest observed number of replicas? What if the user has changed the value of `Spec.ReplicasPerNodeGroup` from their original desired value?

echen-98: So I thought this was the approach we agreed on from #20: in the `SdkFind`, `SdkCreate`, and `SdkUpdate` functions, `ko` is copied from `desired` and updated with the latest observed state because `ko` is the object that will be returned as `latest`. At least for `SdkFind`, there's generated code to do the same thing for `Spec` fields which match the name of fields in the `ReplicationGroup` API type.

The proposed solution was to do the same for fields such as `ReplicasPerNodeGroup` which aren't directly returned by the API, but which require some extra processing (and maybe some additional API calls) to get the latest value of. I think we discussed how the ElastiCache API is less than ideal in this sense and that this is the way we would have to compensate for it. If we don't do this then there will be no difference between `latest` and `desired` and the manager's `Update` will not be invoked when it needs to be.

jaypipes: @echen-98 sorry Eric, I guess I wasn't clear in my comments in #20. I was suggesting using the new generic hooks framework and *only* doing this for the `sdk_read_one_post_set_output` hook. The way you have things here, the `customSetOutput` stuff is being fired for Describe, Create AND Modify, right?

echen-98: @jaypipes Ah okay. Yes it's being called for those three – I was initially looking into using the hooks for ReadOne but was concerned because I thought that only having it for Describe might cause us to miss some things. But Describe gets called each reconciler loop so I think you're right that that's probably the better approach.

Let me follow up with another revision using the hooks for just ReadOne/SdkFind (will probably be awhile since I have various on-calls for the next 2 weeks).

echen-98:
Closing this PR as #29 addresses it.

PR #26, opened 2021-05-18
UserGroup support in ElastiCache

nmvk:
Description of changes:
Support for new resource type UserGroup` in ElastiCache.

Important implementation details 

- Custom modify operation since API takes `--user-ids-to-add` and `--user-ids-to-remove`. (https://docs.aws.amazon.com/cli/latest/reference/elasticache/modify-user-group.html)
- E2E test for basic create update delete
- Updated runtime
- Used new CodeGen so many non UserGroup files are updated.

```
e2e/tests/test_user_group.py::TestUserGroup::test_user_group_create_update 
[gw1] [100%] PASSED e2e/tests/test_user_group.py::TestUserGroup::test_user_group_create_update 

================================================================================== 1 passed in 153.10s (0:02:33) ===================================================================================
```


pkg/resource/user_group/custom_set_output.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+func (rm *resourceManager) CustomDescribeUserGroupsSetOutput(
+	ctx context.Context,
+	r *resource,
+	resp *elasticache.DescribeUserGroupsOutput,
+	ko *svcapitypes.UserGroup,
+) (*svcapitypes.UserGroup, error) {
+	if len(resp.UserGroups) == 0 {
+		return ko, nil
```
kumargauravsharma, line 31: This should be an error scenario

pkg/resource/user_group/custom_update_api.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+
+			input, err := rm.newUpdateRequestPayload(ctx, desired)
+
+			if err != nil {
+				return nil, err
+			}
+
+			input.SetUserIdsToAdd(rm.getUserIdsToBeCreatedOrRemoved(requiredUserIdsMap))
+			input.SetUserIdsToRemove(rm.getUserIdsToBeCreatedOrRemoved(existingUserIdsMap))
```
kumargauravsharma, line 67: method `getUserIdsToBeCreatedOrRemoved` acts as filter.
The code can be made simpler, by removing this method and having two maps populated during the for loop above.

pkg/resource/user_group/custom_update_api.go
```diff
@@ -0,0 +1,182 @@
...
+	latest *resource,
+	delta *ackcompare.Delta,
+) (*resource, error) {
+
+	// Check for the user group status
+	if latest.ko.Status.Status == nil || *latest.ko.Status.Status != "active" {
+		return nil, requeue.NeededAfter(
+			errors.New("user group can not be modified, it is not in 'active' state"),
+			requeue.DefaultRequeueAfterDuration)
+	}
```
kumargauravsharma, line 39: Also include `PendingChanges` check such that if there are pending changes, then defer this update call.

pkg/resource/user_group/custom_update_api.go
```diff
@@ -0,0 +1,182 @@
...
+	for _, userId := range userIds {
+		userIdsMap[*userId] = true
+	}
+
+	return userIdsMap
+}
+
+// getUserIdsToBeCreatedOrRemoved returns list of user ids which have value set as true.
+// Used to compute what user ids we want to add or remove.
+func (rm *resourceManager) getUserIdsToBeCreatedOrRemoved(userIdMap map[string]bool) []*string {
```
kumargauravsharma, line 157: This method can be removed.

test/e2e/tests/test_user_group.py
```diff
@@ -0,0 +1,96 @@
...
+            CRD_GROUP, CRD_VERSION, RESOURCE_PLURAL, user_group_name, namespace="default")
+        _ = k8s.create_custom_resource(reference, usergroup)
+        resource = k8s.wait_resource_consumed_by_controller(reference, wait_periods=10)
+        assert resource is not None
+        return (reference, resource)
+
+    return _make_user_group
+
+@pytest.fixture(scope="module")
+def make_non_default_user():
```
echen-98, line 46: I would suggest placing the creation of any dependent test resources in `service_bootstrap.py`, and their cleanup in `service_cleanup.py`. Not required by any means but it would be nice for consistency with the other tests

test/e2e/tests/test_user_group.py
```diff
@@ -0,0 +1,96 @@
...
+        update_dict = {
+            "USER_GROUP_ID": reference.name,
+            "USER": make_non_default_user
+        }
+
+        # Update the user group
+        updated_spec = load_elasticache_resource(
+            "usergroup_additional_user", additional_replacements=update_dict)
+
+        k8s.patch_custom_resource(reference, updated_spec)
```
echen-98, line 93: I think you may be able to remove the call to `load_elasticache_resource` by doing something like the below:

```
patch = {"spec": {"userIDs": ['default', make_non_default_user]}}
k8s.patch_custom_resource(reference, patch)
```

or just combine those into one line

test/e2e/tests/test_user_group.py
```diff
@@ -0,0 +1,96 @@
...
+
+        assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "True", wait_periods=15)
```
echen-98, line 95: Can you add an assertion that the new user is confirmed to be in the user group?

test/e2e/resources/usergroup_additional_user.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: elasticache.services.k8s.aws/v1alpha1
```
echen-98, line 1: If you are able to patch the resource as indicated in the comment below, we may be able to remove this file

pkg/resource/user_group/custom_update_api.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+
+			resp, respErr := rm.sdkapi.ModifyUserGroupWithContext(ctx, input)
+			rm.metrics.RecordAPICall("UPDATE", "ModifyUserGroup", respErr)
+			if respErr != nil {
+				return nil, respErr
+			}
+			// Merge in the information we read from the API call above to the copy of
+			// the original Kubernetes object we passed to the function
+			ko := desired.ko.DeepCopy()
```
echen-98, line 99: Is the below block of code same as the generated code in `sdk.go`? If so, we may be able to avoid doing this by placing the above code where you add/remove Users in a `post_build_request` hook, and then just defer to the generated code afterwards

pkg/resource/user_group/custom_update_api.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// Check for the user group status
+	if latest.ko.Status.Status == nil || *latest.ko.Status.Status != "active" {
+		return nil, requeue.NeededAfter(
+			errors.New("user group can not be modified, it is not in 'active' state"),
+			requeue.DefaultRequeueAfterDuration)
+	}
+
+	for _, diff := range delta.Differences {
+		if diff.Path.Contains("Spec.UserIDs") {
```
echen-98, line 42: I may have seen this comment elsewhere but I think this can be replaced by `delta.DifferentAt("Spec.UserIDs")`?

echen-98 (commented):
Looks good in general. I think we might be able to simplify the custom code a little. Other than that, I suggested some small changes with the e2e tests

pkg/resource/user_group/custom_update_api.go
echen-98: Actually I think I was incorrect above. I think neither of the approaches are currently functional since `Path.Contains` loops through the parts and looks for an exact match with the subject string. The issue is that one part is "Spec" and one part is "UserIDs" but there is no part named "Spec.UserIDs".

The e2e test may have passed since there is no assertion that the new user is actually contained in the user group.

echen-98: https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/compare/delta.go#L28-L35

The current implementation of `delta.DifferentAt` is potentially problematic IMO. Changing the above code to `delta.DifferentAt("UserIDs")` may make it function, but the risk is that there could some other field (e.g. in `status`, `metadata`) or sub-fields also named `UserIDs` that would fall under this call, not just the field of interest in `spec`.

One possible solution would be to split the supplied subject string and only return true if the resulting slice matches `diff.Path.parts` exactly (in length, content, and order). This seems closer to the intention conveyed in the comment above `DifferentAt`. Relying on `Path.Contains` may be too imprecise.

@jaypipes could you give your input on this issue?

pkg/resource/user_group/custom_update_api.go
```diff
@@ -0,0 +1,182 @@
...
+
+	// Check for the user group status
+	if latest.ko.Status.Status == nil || *latest.ko.Status.Status != "active" {
+		return nil, requeue.NeededAfter(
+			errors.New("user group can not be modified, it is not in 'active' state"),
+			requeue.DefaultRequeueAfterDuration)
+	}
+
+	for _, diff := range delta.Differences {
+		if diff.Path.Contains("UserIDs") && diff.Path.Contains("Spec") {
```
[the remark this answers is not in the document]
A-Hilaly: Currently this is not possible. I just opened https://github.com/aws-controllers-k8s/community/issues/808

pkg/resource/user_group/custom_update_api.go
echen-98: see Amine's comment above (Issue 808)

jaypipes: > @jaypipes could you give your input on this issue?

@echen-98 no need for me to comment on this now that https://github.com/aws-controllers-k8s/runtime/pull/20 is merged, right?

echen-98: @jaypipes correct

pkg/resource/user_group/custom_update_api.go
nmvk: Generated code did not seem to be concise and have additional statements which did not seem to be required thus I created a completely custom update method. I would keep this same for now will again revisit later to see if this can be improved.

test/e2e/tests/test_user_group.py
nmvk: I updated it to assert spec length and status that was probably the best to do

test/e2e/tests/test_usergroup.py
```diff
@@ -0,0 +1,93 @@
...
+        },
+        "kind" : KIND_NAME,
+        "spec": {
+            "engine": "redis",
+            "userGroupID": user_group_name,
+            "userIDs": [
+                "default"
+            ]
+        }
+    }
```
kumargauravsharma, line 53: This can be moved to `/test/e2e/resources` as `UserGroup.json` and referenced here using `load_elasticache_resource` from e2e.

test/e2e/tests/test_usergroup.py
```diff
@@ -0,0 +1,93 @@
...
+            "kind": KIND_NAME,
+            "spec": {
+                "engine": "redis",
+                "userGroupID": reference.name,
+                "userIDs": [
+                    "default",
+                    bootstrap_resources.NonDefaultUser
+                ]
+            }
+        }
```
kumargauravsharma, line 86: Similar to previous comment: This can be moved to `/test/e2e/resources` as `UserGroup.json` and referenced here using `load_elasticache_resource` from e2e.
And then read reference can be updated to have user Id added to it.

test/e2e/tests/test_usergroup.py
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    # Create new user group
+    _ = k8s.create_custom_resource(reference, user_group)
+    resource = k8s.wait_resource_consumed_by_controller(reference, wait_periods=10)
+
+    yield reference, resource
+
+    # Teardown
+    k8s.delete_custom_resource(reference)
```
A-Hilaly, line 62:
```suggestion
    _, deleted = k8s.delete_custom_resource(reference)
    assert deleted is True
```

A-Hilaly (commented):
LGTM overall :+1: 
I Left few comments below

pkg/resource/user_group/custom_update_api.go
echen-98: Sounds good. Before, I was aware of `custom_implementation` in the generator config but didn't know you could replace the entire `sdkUpdate` using `update_operation.custom_method_name`.

pkg/resource/user_group/custom_update_api.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// Check for the user group status
+	if latest.ko.Status.Status == nil || *latest.ko.Status.Status != "active" {
+		return nil, requeue.NeededAfter(
+			errors.New("user group can not be modified, it is not in 'active' state"),
+			requeue.DefaultRequeueAfterDuration)
+	}
+
+	for _, diff := range delta.Differences {
+		if diff.Path.Contains("UserIDs") {
```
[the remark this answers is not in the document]
nmvk: It is `delta.DifferentAt` and not `diff.DifferentAt` so this would need to be same.

pkg/resource/user_group/custom_update_api.go
```diff
@@ -0,0 +1,185 @@
...
+				var userIdsToRemove []*string
+
+				for userId, include := range existingUserIdsMap {
+					if include {
+						userIdsToRemove = append(userIdsToRemove, &userId)
+					}
+				}
+
+				input.SetUserIdsToRemove(userIdsToRemove)
+			}
```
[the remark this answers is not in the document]
nmvk: Thanks @A-Hilaly I agree with moving this to a common package/file, right now going to leave it like this as-is to avoid any readability issues.

pkg/resource/user_group/custom_update_api.go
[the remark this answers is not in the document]
A-Hilaly: Yup `delta.DifferentAt` ++

pkg/resource/user_group/custom_set_output.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (rm *resourceManager) CustomDescribeUserGroupsSetOutput(
+	ctx context.Context,
+	r *resource,
+	resp *elasticache.DescribeUserGroupsOutput,
+	ko *svcapitypes.UserGroup,
+) (*svcapitypes.UserGroup, error) {
+	if len(resp.UserGroups) == 0 {
+		return nil, ackerr.NotFound
+	}
```
jaypipes, line 33: This isn't necessary. See lines 140-142 of sdk.go and you will notice that CustomDescribeUserGroupsSetOutput will never be called if `len(resp.UserGroups) == 0`.

pkg/resource/user_group/custom_set_output.go
```diff
@@ -0,0 +1,97 @@
...
+		ko.Spec.UserIDs = userIds
+	}
+
+	if engine != nil {
+		ko.Spec.Engine = engine
+	}
+
+	if ko.Status.Conditions == nil {
+		ko.Status.Conditions = []*ackv1alpha1.Condition{}
+	}
```
jaypipes, line 71: The above is not necessary. Line 144 and line 237 in sdk.go will have already done this for you.

jaypipes (commented):
The custom code in elasticache-controller can/should be replaced with the simpler generic hooks system over time (to reduce the amount of custom code that you need to maintain). This is good for now (though I left a couple suggestions for code you can simply remove if you push another revision to this PR).

Thanks @nmvk! :)

kumargauravsharma (approved):

echen-98 (approved):

PR #29, opened 2021-05-24
Update spec fields in ReadOne post_set_output hook

echen-98:
Description of changes:
Related to https://github.com/aws-controllers-k8s/elasticache-controller/pull/22

To recap, in order to trigger the `Update` code, there needs to be a Spec difference between `desired` and `latest`. The generated `ReadOne` code often doesn't update `latest.Spec` with the latest observed state, since determining the latest observed state often requires some processing or additional API calls. There are many fields this applies to, but this PR starts with `ReplicasPerNodeGroup` and `EngineVersion`

### DIFF
--- a/pkg/resource/replication_group/custom_update_api.go
+++ b/pkg/resource/replication_group/custom_update_api.go
@@ -134,7 +134,7 @@ func (rm *resourceManager) modifyReplicationGroup(
 
 	// SecurityGroupIds, EngineVersion
 	if rm.securityGroupIdsDiffer(desired, latest, latestCacheCluster) ||
-		rm.engineVersionsDiffer(desired, latestCacheCluster) {
+		rm.engineVersionsDiffer(desired, latest) {
 		input := rm.newModifyReplicationGroupRequestPayload(desired, latest, latestCacheCluster)
 		resp, respErr := rm.sdkapi.ModifyReplicationGroupWithContext(ctx, input)
 		rm.metrics.RecordAPICall("UPDATE", "ModifyReplicationGroup", respErr)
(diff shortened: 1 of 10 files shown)


pkg/resource/replication_group/post_set_output.go
```diff
@@ -0,0 +1,70 @@
...
+
+/*
+	To be called in sdkFind, this function updates the replication group's Spec fields with the latest observed state
+	This requires extra processing of the API response as well as additional API calls, and is necessary because
+	sdkFind does not update many of these Spec fields by default.
+*/
+func (rm *resourceManager) updateSpecFields(
+	ctx context.Context,
+	respRG *elasticache.ReplicationGroup,
+	ko *svcapitypes.ReplicationGroup,
```
jaypipes, line 31: recommend just passing a `*resource` and do the `ko` unpacking inside the function. That way you don't need to import the `svcapitypes` package.

pkg/resource/replication_group/custom_update_api.go
```diff
@@ -127,14 +127,14 @@ func (rm *resourceManager) modifyReplicationGroup(
 	}
 
 	// Get details using describe cache cluster to compute diff
-	latestCacheCluster, err := rm.describeCacheCluster(ctx, latest)
+	latestCacheCluster, err := rm.describeCacheCluster(ctx, latest.ko)
```
jaypipes, line 130: This change seems unnecessary... why not just keep the describeCacheCluster function signature the same and "unpack" the `ko` field from within the function body?

pkg/resource/replication_group/custom_update_api.go
```diff
@@ -711,27 +676,25 @@ func (rm *resourceManager) newModifyReplicationGroupRequestPayload(
...
-desired: the resource representing the desired state
-latestCacheCluster: a CacheCluster object representing one cache node in the replication group, which
-	reveals the currently used engine version for the entire replication group
+from the latest observed engine version, and false if they differ or if
+the desired EngineVersion is nil
 */
 func (rm *resourceManager) engineVersionsDiffer(
 	desired *resource,
-	latestCacheCluster *svcsdk.CacheCluster,
+	latest *resource,
```
jaypipes, line 684: :+1: the above is how all these function signatures should be structured.

pkg/resource/replication_group/post_set_output.go
```diff
@@ -0,0 +1,70 @@
...
+	latestCacheCluster, err := rm.describeCacheCluster(ctx, ko)
+	if err == nil && latestCacheCluster != nil {
+		setEngineVersion(latestCacheCluster, ko)
+	}
+}
+
+// if ReplicasPerNodeGroup was given in desired.Spec, update ko.Spec with the latest observed value
+func setReplicasPerNodeGroup(
+	respRG *elasticache.ReplicationGroup,
+	ko *svcapitypes.ReplicationGroup,
```
jaypipes, line 49: same comment as above.

pkg/resource/replication_group/post_set_output.go
```diff
@@ -0,0 +1,70 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package replication_group
+
+import (
+	"context"
+
+	svcapitypes "github.com/aws-controllers-k8s/elasticache-controller/apis/v1alpha1"
+	"github.com/aws/aws-sdk-go/service/elasticache"
```
jaypipes, line 20: recommend aliasing this package as `svcsdk` as we do elsewhere in all the `pkg/resource` templates:

https://github.com/aws-controllers-k8s/code-generator/blob/cf8f94fd9835b500cbf736fcea48efe0ca37f8a1/templates/pkg/resource/sdk.go.tpl#L13

It's easier to always use this alias and by using the alias, you allow readers of the code to imagine how this code might be used in other service-specific controllers.

jaypipes (changes requested):
Couple minor things, please see inline.

echen-98:
@jaypipes thanks for the comments. Changes made as requested

jaypipes (approved):
:+1:

kumargauravsharma (approved):

PR #33, opened 2021-06-14
fix Prow e2e test run failures

echen-98:
Description of changes:

Address e2e test failures:
- `test_rg_input_coverage` was failing because the type of the `authToken` field has changed since the test was written
- `test_rg_auth_token` was failing because the controller does not have permissions to list secrets. For now, we are skipping this test. See issue https://github.com/aws-controllers-k8s/community/issues/745
- `test_rg_cmd_fromsnapshot` succeeds locally but failed in the last Prow run. This was likely due to a region mismatch (tests run in us-west-2, input yamls specify us-east-1 AZs)

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -19,6 +19,7 @@ from acktest import k8s
 
 def pytest_addoption(parser):
     parser.addoption("--runslow", action="store_true", default=False, help="run slow tests")
+    parser.addoption("--runblocked", action="store_true", default=False, help="run blocked tests")
 
 
 def pytest_configure(config):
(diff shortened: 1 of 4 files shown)


jaypipes (approved):
:+1:

echen-98:
Prow job seems to be having issues with replication group tests (even though they succeed locally). I'll start a new run to see if the results are reliable and then attempt to further investigate/reproduce the issue.

/test elasticache-kind-e2e

kumargauravsharma (approved):

echen-98:
Adjusted AZs, running again:

/ok-to-test
/test elasticache-kind-e2e
/test elasticache-unit-test

echen-98:
Actual tests are now passing (see prow output above), but we are facing a credential timeout upon teardown. We can merge this PR and I will follow up with another to parallelize the replication group tests so they don't run sequentially.

PR #34, opened 2021-06-15
Delete replication group to wait for delete completion

kumargauravsharma:
Issue #, if available:

Replication group delete takes some time to delete while `kubectl delete -f replication_group.yaml` removes the custom resource record from cluster.

Description of changes:

This PR adds logic inside `sdk::sdkDelete()` to requeue reconcile while replication group is deleting.

### DIFF
--- a/generator.yaml
+++ b/generator.yaml
@@ -58,6 +58,10 @@ resources:
     hooks:
       sdk_read_many_post_set_output:
         code: "rm.updateSpecFields(ctx, resp.ReplicationGroups[0], &resource{ko})"
+      sdk_delete_pre_build_request:
+        template_path: hooks/sdk_delete_pre_build_request.go.tpl
+      sdk_delete_post_request:
+        template_path: hooks/sdk_delete_post_request.go.tpl
   Snapshot:
     update_conditions_custom_method_name: CustomUpdateConditions
     exceptions:
--- a/templates/hooks/sdk_delete_post_request.go.tpl
+++ b/templates/hooks/sdk_delete_post_request.go.tpl
@@ -1,0 +1,8 @@
+	if respErr == nil {
+		if foundResource, err := rm.sdkFind(ctx, r); err != ackerr.NotFound {
+			if isDeleting(foundResource) {
+				return requeueWaitWhileDeleting
+			}
+			return err
+		}
+	}
--- a/templates/hooks/sdk_delete_pre_build_request.go.tpl
+++ b/templates/hooks/sdk_delete_pre_build_request.go.tpl
@@ -1,0 +1,4 @@
+	if isDeleting(r) {
+		return requeueWaitWhileDeleting
+	}
+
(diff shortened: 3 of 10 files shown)


templates/hooks/sdk_delete_pre_build_request.go.tpl
```diff
@@ -0,0 +1,4 @@
+	if isDeleting(r) {
+		return requeueWaitWhileDeleting
```
jaypipes, line 2: Did you need to `git add` a `conditions.go` file? :) I don't see where `requeueWaitWhileDeleting` is defined?

jaypipes (changes requested):
Looking great @kumargauravsharma  but I think you may have forgotten to git add a file...? :)

templates/hooks/sdk_delete_pre_build_request.go.tpl
kumargauravsharma: `requeueWaitWhileDeleting` is defined in `hooks.go`

generator.yaml
```diff
@@ -55,6 +55,11 @@ resources:
           path: Events
       AuthToken:
         is_secret: true
+    hooks:
```
A-Hilaly, line 58: Not sure but maybe you need a `sdk_find_post_set_output` hook as well? (similar to `dk_delete_pre_build_request.go.tpl`)

generator.yaml
```diff
@@ -55,6 +55,11 @@ resources:
...
+      sdk_delete_pre_build_request:
+        template_path: hooks/sdk_delete_pre_build_request.go.tpl
+      sdk_delete_post_request:
+          template_path: hooks/sdk_delete_post_request.go.tpl
```
A-Hilaly, line 62:
```suggestion
        template_path: hooks/sdk_delete_post_request.go.tpl
```

kumargauravsharma: Removed the extra space.

generator.yaml
kumargauravsharma: It was not needed.
I have a upcoming separate PR to patch resource which is being deleted, so that kubectl describe provides latest state of the resource as it gets deleted.
But that is independent of this PR.

templates/hooks/sdk_delete_pre_build_request.go.tpl
jaypipes: doh! sorry, not sure how I missed that :)

jaypipes (approved):

echen-98 (approved):

nmvk (approved):

templates/hooks/sdk_delete_post_request.go.tpl
```diff
@@ -0,0 +1,8 @@
+	if respErr == nil {
+		if foundResource, err := rm.sdkFind(ctx, r); err != ackerr.NotFound {
```
jaypipes, line 2: If a DELETE operation on ReplicationGroup has returned no error, doesn't this mean that the resource will *always* be in a `deleting` status? Is it necessary to do another sdkFind() operation here?

Further, if a DELETE operation on ReplicationGroup returns no error, is there any chance that the ReplicationGroup can end up *not* being deleted in the backend Elasticache service? If there is no chance of this, is it really necessary to do anything here other than return nil and let the elasticache-controller reconcile the resource and remove its finalizer in `cleanup()` and get rid of the CR on the k8s side?

PR #35, opened 2021-06-16
address regression with User code, fix RG deletion test

echen-98:
Description of changes:

- Change User code to be compatible with https://github.com/aws-controllers-k8s/runtime/pull/20
- Fix RG deletion test (pytest was complaining about a direct call to a fixture)

Local e2e test run:
```
============================================================================ test session starts =============================================================================
platform linux -- Python 3.8.10, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /elasticache-controller/tests/e2e
plugins: xdist-2.2.0, forked-1.3.0
[gw0] Python 3.8.10 (default, May  4 2021, 19:07:02)  -- [GCC 10.2.1 20201203]
[gw1] Python 3.8.10 (default, May  4 2021, 19:07:02)  -- [GCC 10.2.1 20201203]
[gw2] Python 3.8.10 (default, May  4 2021, 19:07:02)  -- [GCC 10.2.1 20201203]
[gw3] Python 3.8.10 (default, May  4 2021, 19:07:02)  -- [GCC 10.2.1 20201203]
[gw4] Python 3.8.10 (default, May  4 2021, 19:07:02)  -- [GCC 10.2.1 20201203]
[gw5] Python 3.8.10 (default, May  4 2021, 19:07:02)  -- [GCC 10.2.1 20201203]
gw0 [9] / gw1 [9] / gw2 [9] / gw3 [9] / gw4 [9] / gw5 [9]
scheduling tests via LoadScheduling

tests/test_replicationgroup.py::TestReplicationGroup::test_rg_input_coverage 
tests/test_replicationgroup.py::TestReplicationGroup::test_rg_cmd_update 
tests/test_replicationgroup.py::TestReplicationGroup::test_rg_delete 
tests/test_replicationgroup_largecluster.py::TestReplicationGroupLargeCluster::test_rg_largecluster 
tests/test_replicationgroup.py::TestReplicationGroup::test_rg_cmd_fromsnapshot 
tests/test_replicationgroup.py::TestReplicationGroup::test_rg_auth_token 
[gw5] [ 11%] SKIPPED tests/test_replicationgroup_largecluster.py::TestReplicationGroupLargeCluster::test_rg_largecluster 
[gw3] [ 22%] SKIPPED tests/test_replicationgroup.py::TestReplicationGroup::test_rg_auth_token x
[gw1] [ 33%] PASSED tests/test_replicationgroup.py::TestReplicationGroup::test_rg_cmd_fromsnapshot 
[gw0] [ 44%] PASSED tests/test_replicationgroup.py::TestReplicationGroup::test_rg_input_coverage 
[gw4] [ 55%] PASSED tests/test_replicationgroup.py::TestReplicationGroup::test_rg_delete 
tests/test_user.py::TestUser::test_CRUD 
[gw1] [ 66%] PASSED tests/test_user.py::TestUser::test_CRUD 
tests/test_snapshot.py::TestSnapshot::test_snapshot_kms 
[gw2] [ 77%] PASSED tests/test_replicationgroup.py::TestReplicationGroup::test_rg_cmd_update 
[gw0] [ 88%] PASSED tests/test_snapshot.py::TestSnapshot::test_snapshot_kms 
tests/test_usergroup.py::TestUserGroup::test_user_group_create_update 
[gw2] [100%] PASSED tests/test_usergroup.py::TestUserGroup::test_user_group_create_update 

================================================================= 7 passed, 2 skipped in 1688.28s (0:28:08) ==================================================================

```

### DIFF
--- a/pkg/resource/user/delta_util.go
+++ b/pkg/resource/user/delta_util.go
@@ -23,12 +23,10 @@ func filterDelta(
 ) {
 	// the returned AccessString can be different than the specified one; as long as the last requested AccessString
 	// matches the currently desired one, remove this difference from the delta
-	//TODO: revert this call to Spec.AccessString once we have a new implementation of it
-	if delta.DifferentAt("AccessString") {
+	if delta.DifferentAt("Spec.AccessString") {
 		if *desired.ko.Spec.AccessString == *desired.ko.Status.LastRequestedAccessString {
 
-			//TODO: revert the call to Spec.AccessString once removeFromDelta implementation changes
-			removeFromDelta(delta, "AccessString")
+			removeFromDelta(delta, "Spec.AccessString")
 		}
 	}
 }
@@ -43,7 +41,7 @@ func removeFromDelta(
 	differences := delta.Differences
 
 	// identify index of Difference to remove
-	//TODO: change once we get a Path.Equals or similar method
+	//TODO: this could require a stricter Path.Equals down the road
 	var i *int = nil
 	for j, diff := range differences {
 		if diff.Path.Contains(subject) {
--- a/test/e2e/tests/test_replicationgroup.py
+++ b/test/e2e/tests/test_replicationgroup.py
@@ -154,6 +154,24 @@ def rg_cmd_update(rg_cmd_update_input, make_replication_group, rg_deletion_waite
     rg_deletion_waiter.wait(ReplicationGroupId=input_dict["RG_ID"])
 
 
+@pytest.fixture(scope="module")
+def rg_deletion_input(make_rg_name):
+    return {
+        "RG_ID": make_rg_name("rg-delete"),
+        "ENGINE_VERSION": "6.x",
+        "NUM_NODE_GROUPS": "1",
+        "REPLICAS_PER_NODE_GROUP": "1"
+    }
+
+
+@pytest.fixture(scope="module")
+def rg_deletion(rg_deletion_input, make_replication_group):
+    input_dict = rg_deletion_input
+
+    (reference, resource) = make_replication_group("replicationgroup_cmd_update", input_dict, input_dict["RG_ID"])
+    return (reference, resource)  # no teardown, as the teardown is part of the actual test
+
+
 @service_marker
 class TestReplicationGroup:
 
@@ -229,9 +247,8 @@ class TestReplicationGroup:
         k8s.patch_custom_resource(reference, updated_spec)
         assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "True", wait_periods=30)
 
-    def test_rg_delete(self, rg_cmd_update_input, rg_deletion_waiter):
-        input_dict = rg_cmd_update_input
-        (reference, _) = make_replication_group("replicationgroup_cmd_update", input_dict, input_dict["RG_ID"])
+    def test_rg_deletion(self, rg_deletion_input, rg_deletion, rg_deletion_waiter):
+        (reference, _) = rg_deletion
         assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "True", wait_periods=30)
 
         # assertions after initial creation
@@ -244,7 +261,7 @@ class TestReplicationGroup:
 
         resource = k8s.get_resource(reference)
         assert resource['metadata']['deletionTimestamp'] is not None
-        # uncomment when reconciler->cleanup() invokes patchResource()
+        # TODO: uncomment when reconciler->cleanup() invokes patchResource()
         # assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "False", wait_periods=1)
 
-        rg_deletion_waiter.wait(ReplicationGroupId=input_dict["RG_ID"])
+        rg_deletion_waiter.wait(ReplicationGroupId=rg_deletion_input["RG_ID"])
--- a/pkg/resource/user/post_build_request.go
+++ b/pkg/resource/user/post_build_request.go
@@ -26,13 +26,11 @@ func (rm *resourceManager) populateUpdatePayload(
 	r *resource,
 	delta *ackcompare.Delta,
 ) {
-	//TODO: change all calls to DifferentAt to include full path once the implementation is clarified
-
-	if delta.DifferentAt("AccessString") && r.ko.Spec.AccessString != nil {
+	if delta.DifferentAt("Spec.AccessString") && r.ko.Spec.AccessString != nil {
 		input.AccessString = r.ko.Spec.AccessString
 	}
 
-	if delta.DifferentAt("NoPasswordRequired") && r.ko.Spec.NoPasswordRequired != nil {
+	if delta.DifferentAt("Spec.NoPasswordRequired") && r.ko.Spec.NoPasswordRequired != nil {
 		input.NoPasswordRequired = r.ko.Spec.NoPasswordRequired
 	}
 


nmvk (approved):
LGTM

pkg/resource/user/delta_util.go
```diff
@@ -23,12 +23,10 @@ func filterDelta(
...
 	// the returned AccessString can be different than the specified one; as long as the last requested AccessString
 	// matches the currently desired one, remove this difference from the delta
-	//TODO: revert this call to Spec.AccessString once we have a new implementation of it
-	if delta.DifferentAt("AccessString") {
+	if delta.DifferentAt("Spec.AccessString") {
 		if *desired.ko.Spec.AccessString == *desired.ko.Status.LastRequestedAccessString {
 
-			//TODO: revert the call to Spec.AccessString once removeFromDelta implementation changes
-			removeFromDelta(delta, "AccessString")
+			removeFromDelta(delta, "Spec.AccessString")
```
jaypipes, line 29: Why do you have multiple "AccessString" fields in the `Status` struct?

https://github.com/aws-controllers-k8s/elasticache-controller/blob/e0d2cb95702866ec24b7147a9b52d2bea8f5ba06/generator.yaml#L122-L131

kumargauravsharma (approved):

test/e2e/tests/test_replicationgroup.py
```diff
@@ -244,7 +261,7 @@ def test_rg_delete(self, rg_cmd_update_input, rg_deletion_waiter):
 
         resource = k8s.get_resource(reference)
         assert resource['metadata']['deletionTimestamp'] is not None
-        # uncomment when reconciler->cleanup() invokes patchResource()
+        # TODO: uncomment when reconciler->cleanup() invokes patchResource()
```
jaypipes, line 264: reconciler->cleanup() does not invoke `patchResource()` because it finalizes the deletion of the resource from Kubernetes.

pkg/resource/user/delta_util.go
echen-98: The access string provided by the ElastiCache API differs from the specified access string (there's some behind-the-scenes expansion). 

Because there's no straightforward way to translate between the two representations, adding a state-keeping field `Status.LastRequestedAccessString` allows us to determine when the desired access string has changed. This field is only updated upon a successful create or update (see the `set_output` methods).

Since the generated `delta` code will always detect a difference between `desired.Spec.AccessString` and `latest.Spec.AccessString` there's a hook after that to then remove the difference if `desired.Spec.AccessString` is still equal to `desired.Status.LastRequestAccessString`. See `filterDelta()` in `pkg/resource/user/delta_util.go`.

`Status.ExpandedAccessString` is purely informational and was added because that's what the user would see if they were directly using the ElastiCache API.

test/e2e/tests/test_replicationgroup.py
echen-98: That discussion is out of scope of this PR, I just added a `TODO` here since it felt like the comment needed one.

More details:
https://github.com/aws-controllers-k8s/runtime/pull/21
https://github.com/aws-controllers-k8s/elasticache-controller/pull/34

@kumargauravsharma can elaborate if needed.

pkg/resource/user/delta_util.go
jaypipes: @echen-98 why aren't you simply updating `Spec.AccessString` to the actual expanded value upon return from CreateUser and just get rid of the two other fields in the Status?

echen-98: @jaypipes do you mean updating `desired.Spec.AccessString` or `latest.Spec.AccessString`? We're already doing the latter, and unless I'm missing something big, I thought the former wasn't possible?

echen-98: i.e. I thought `patchResource()` only patched `Status` because `desired.Spec` should not be modified

jaypipes: > @jaypipes do you mean updating `desired.Spec.AccessString` or `latest.Spec.AccessString`? We're already doing the latter, and unless I'm missing something big, I thought the former wasn't possible?

`latest.Spec.AccessString`

If you're already doing that, why do you need multiple copies of that value in `Status.LastRequestedAccessString` and `Status.ExpandedAccessString`?

echen-98: @jaypipes

`desired.Spec.AccessString`: `on ~app::* -@all +@read`
`latest.Spec.AccessString`: `on ~app::* -@all +@read +@hash +@bitmap +@geo -setbit -bitfield -hset -hsetnx -hmset -hincrby -hincrbyfloat -hdel -bitop -geoadd -georadius -georadiusbymember`

There is always a `Spec` difference in `AccessString`. Without any custom code, this path gets added to the `delta`, and an Update is always invoked. So the default behavior of the generated code is to make Update calls in an infinite loop.

Even if we introduce custom update code, we have no way of telling whether we should update or not (since as I said above, we can't translate between the two representations). So this leaves us no better than the generated code.

The only way to tell if the desired state has changed is to add a state-keeping field such as `Status.LastRequestedAccessString`. Although this path will always be added to the delta, this allows us to remove it from the delta if there has been no change, and therefore not invoke an Update at all. If there is a difference, then the path is left in the delta and the resulting Update call is actually meaningful.

As for `Status.ExpandedAccessString`, `latest.Spec` is not visible to the user, so the only way to make it visible is to make it a `Status` field. This is actually a larger issue with other resources as well.

Let me know if my explanation is lacking anywhere and I can try to elaborate further.

jaypipes: > @jaypipes
> 
> `desired.Spec.AccessString`: `on ~app::* -@all +@read`
> `latest.Spec.AccessString`: `on ~app::* -@all +@read +@hash +@bitmap +@geo -setbit -bitfield -hset -hsetnx -hmset -hincrby -hincrbyfloat -hdel -bitop -geoadd -georadius -georadiusbymember`
> 
> There is always a `Spec` difference in `AccessString`. Without any custom code, this path gets added to the `delta`, and an Update is always invoked. So the default behavior of the generated code is to make (mostly useless) Update calls in an infinite loop.

If we were saving `Spec.AccessString` as the value returned by the server after a CreateUser or UpdateUser call, there would be no infinite loop.

echen-98: So in the Create case, it was my impression that in the reconciler, after a resource is found to not exist and a Create is called, then a ReadOne will be called next. If this ReadOne populates `latest.Spec.AccessString` with the expanded access string, there will be a `Spec` difference (since `desired.Spec` has the user-requested access string). And as long as there is a `Spec` difference, then `Update` will be invoked.

What I'm trying to say is it doesn't matter whether we save the server-returned value in `latest.Spec.AccessString` in a Create or Update call (especially since `latest.Spec` is thrown away upon patching in these two cases). The infinite loop depends solely upon what happens in `ReadOne`, and `ReadOne` will always populate `latest.Spec.AccessString` with a value different from `desired.Spec.AccessString`.

echen-98: Sorry, I should add that there may not be an infinite loop (that was the initial behavior when I submitted my first iteration of the User code before the split repos, and there have been several runtime changes since then). However, this would still cause an unnecessary Update call after the creation of the resource, because the ReadOne right after the Create would pick up the `Spec` difference. 

Regardless, the point is that we need to store the last requested access string. Otherwise we won't know when the desired access string has changed. If the user updates another field other than the access string, we want to know whether or not to add the access string to the Update payload (I don't think it's best practice to add a field unless it actually has changed).

This is the approach that we decided on before the split repos, and we decided that we should keep it once the User resource was brought back into scope. It works, and if it's sub-optimal, that just reflects that we sometimes have to do sub-optimal things when dealing with the ElastiCache API.

The real point of this discussion is the more general question of: how do you handle the case where an API returns a value different from the one requested? I think that's a very important question to answer, but I also think that this PR is not the place to have this discussion. The point of this PR was to finally have succeeding Prow runs, and to continue to block it until we have a clear answer to that question doesn't make sense to me.

Also, even if we were to settle on a different solution, that would be in a follow-up PR anyway.

https://github.com/echen-98/ack-community/commit/f0590646edc94fff9e8d58dd0ecca68b86b4f361
https://github.com/aws-controllers-k8s/elasticache-controller/pull/32

kumargauravsharma: I agree with @echen-98
We shall let this PR go through and have different issues to track discussion on Spec, Status population and Delta computation and resource patching related aspects for special fields (readonly, server modifiable, not specified in spec but server provided defaults etc.).

kumargauravsharma: @jaypipes please suggest if its fine to merge this PR in its current state.

test/e2e/tests/test_replicationgroup.py
```diff
@@ -154,6 +154,24 @@ def rg_cmd_update(rg_cmd_update_input, make_replication_group, rg_deletion_waite
...
+    return {
+        "RG_ID": make_rg_name("rg-delete"),
+        "ENGINE_VERSION": "6.x",
+        "NUM_NODE_GROUPS": "1",
+        "REPLICAS_PER_NODE_GROUP": "1"
+    }
+
+
+@pytest.fixture(scope="module")
+def rg_delete(rg_delete_input, make_replication_group):
```
jaypipes, line 168: this function doesn't delete anything...

echen-98: Ok, changed it to the noun form "deletion" to avoid further confusion

jaypipes (approved):
:+1:

echen-98:
Missed some references to renamed fixtures, addressing those and waiting for tests to pass

echen-98:
previous run failed during Docker build step?
/test elasticache-kind-e2e

PR #39, opened 2021-07-01
Enable Passwords field (upon creation)

echen-98:
Description of changes:

Use recent changes in code-generator to allow creation of a User with the `Passwords` field as a slice of k8s secrets. Test included.

Currently, updating the `Passwords` field is not supported (i.e. it's ignored in delta comparison and not placed in the Update payload). This is because in general, we cannot tell when the desired passwords have changed, other than cases where the changes are visible in the `Authentication` field.

Figuring out the update behavior will come later (likely requires changes in common repos).


kumargauravsharma (approved):

nmvk (approved):
LGTM

jaypipes (approved):
:+1: nice work, @echen-98

PR #40, opened 2021-07-10
Declarative integration test framework

kumargauravsharma:
Issue #, if available:

Create, Read, Update, Delete Integration tests across custom resources scenarios contain test code, fixtures that are repetitive across test files. 

Description of changes:

A declarative integration test framework to write integration test as yaml and a reusable common logic for their execution:
* Each test yaml represents an integration test scenario with 1 or many ordered steps.
* Each step specifies a verb (apply, patch, delete) and corresponding expectations to assert
* Scenarios can run in parallel
* Scenarios input yaml can have input replacements (`$SOME_VARIABLE`) which are replaced during test run
* Scenario tear down logic is automatic
* Scenarios can be marked for inclusion/exclusion with custom markers and usecase markers
* Allows custom resource specific customization using `ResourceHelper` auto discovery using `@resource_helper` decorator.
* Uses `session` scoped fixture for service bootstrap resources and input replacements, removing the need for call to `service_bootstrap.py` and `service_cleanup.py` from `test-infra` made [here](https://github.com/aws-controllers-k8s/test-infra/blob/main/scripts/run-tests.sh#L84) and [here](https://github.com/aws-controllers-k8s/test-infra/blob/main/scripts/run-tests.sh#L91).

Declarative integration test framework (` test/declarative_test_fwk/`) is reusable, this PR is reviewed and merged, it can be moved to `test-infra` later.
It can be enhanced further as needed.

Refer `test/e2e/declarative_tests.py`  file in this PR for using declarative integration test framework.
* `test/e2e/scenarios/scenario1.yaml` represent integration test scenario as declarative  yaml  


test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,181 @@
...
+            self.expectations = self.config.get("expect")
+            self.helper = helper.get_resource_helper(self.input_data.get("kind").lower())
+
+    def id(self) -> str:
+        return self.config.get("id", "")
+
+    def description(self) -> str:
+        return self.config.get("description", "")
+
+    def run(self) -> None:
```
kumargauravsharma, line 57: Run logic can be moved to separate file (example: `runner.py`) later.

kumargauravsharma: Added runner.py

RedbackThomson:
Super interesting ideas explored here. Love the idea of moving a lot of the testing to declarative, as most tests run in a fairly consistent CRUD manner. Just have some questions about implementation choices:

- Is there a reason why scenarios are not linked 1:1 with custom resource types? Is this to allow for a scenario to run tests that require multiple CRs?
- Does this implementation allow for server-side checks as well? That is, asserting that the AWS service has properly registered the resource.
- What is the intended effect of the `usecases` marks?
- When running multiple scenarios, how easy is it to debug the output in PyTest? Currently PyTest is very readable, as each test is a separate class and the assertions are very clearly laid out in static Python code.

kumargauravsharma:
Thanks for your inputs @RedbackThomson 

> Is there a reason why scenarios are not linked 1:1 with custom resource types? 
> Is this to allow for a scenario to run tests that require multiple CRs?

In the initial revision of this PR, I considered each step to have custom response details specified in it. It was for flexibility so that scenarios that require multiple CRs, for example pre-req dependency CRs (example: User/UserGroup CR for ReplicationGroup CR), can have steps to create them.
But it also lead to repeated resource reference details (metadata, kind, apiversion etc) in each step even when all the steps are for single resource.
To remove this repetition while keeping the flexibility to specify resource at each step, the current revision of the PR allows `customResourceReference` field at the scenario level.

```
customResourceReference:
  apiVersion: $CRD_GROUP/$CRD_VERSION
  kind: ReplicationGroup
  metadata:
    name: test$RANDOM_SUFFIX
steps:
```
This information is mixed in into each step such that details specified in step takes precedence.
Thus if scenario is for a single resource then there is no need to repeat it in every step, but steps can override it if scenario requires that.

kumargauravsharma:
> Does this implementation allow for server-side checks as well? That is, asserting that the AWS service has properly registered the resource.

Yes, the `ResourceHelper` mechanism allows service controllers to provide resource specific helper implementation. 
By overriding `assert_expectations` service controllers can have logic for server side checks based on step's verb, input data to verb, specified expectations for given custom resource:
```
@helper.resource_helper("ReplicationGroup")
class ReplicationGroupHelper(helper.ResourceHelper):
    def assert_expectations(self, verb: str, input_data: dict, expectations: dict,
                            reference: k8s.CustomResourceReference):
        # default assertions
        super().assert_expectations(verb, input_data, expectations, reference)
        
        # perform custom server side checks based on:
        # verb, input data to verb, expectations for given resource
```

kumargauravsharma:
> What is the intended effect of the usecases marks?

I considered it to allowlist specific custom resource types in a test suite run. This is similar to service marker that we have currently, but its scope is further finegrained to custom resource type.

As single python file is used for all test scenarios, I expect it to provide input configurability to run subset of scenarios from entire test suite.

kumargauravsharma:
> When running multiple scenarios, how easy is it to debug the output in PyTest? Currently PyTest is very readable, as each test is a separate class and the assertions are very clearly laid out in static Python code.


Following help analyze test step failure:

1. Each scenario is executed as individual test with unique test id. The test id includes the test scenario file name:
For example if scenarios are "scenario2.yaml" and "scenario1.yaml" then corresponding test ids are:
```
declarative_tests.py::TestSuite::test_scenario[scenario2.yaml] 
declarative_tests.py::TestSuite::test_scenario[scenario1.yaml] 
```

2. Test framework uses `logging`, when log level is configured (for example at `INFO` level using `--log-cli-level=INFO`), failure details are logged. It provides information about the scenario, step that failed for the test id:
```
declarative_tests.py::TestSuite::test_scenario[scenario2.yaml] 
-------------------------------- live log call ---------------------------------
INFO     root:runner.py:31 Execute: Scenario(id='transitEncryptionEnabled_on_CMD_replication_group_causes_Advisory_condition')
INFO     root:runner.py:75 Execute: Step(id='create_CMD_replication_group_with_transitEncryptionEnabled')
INFO     root:runner.py:115 assert:  Step(id='create_CMD_replication_group_with_transitEncryptionEnabled')
ERROR    root:runner.py:124 AssertionError at Step(id='create_CMD_replication_group_with_transitEncryptionEnabled')
FAILED                                                                   [ 50%]
```

3. Upon test failure, pytest assert introspection provides information about the scenario that failed:
It contains information about `scenario`, and failed step's `expectations`, resource reference `reference`, and the assertion values that failed:
```
declarative_tests.py:90 (TestSuite.test_scenario[scenario2.yaml])
self = <e2e.declarative_tests.TestSuite object at 0x106236eb0>
scenario = Scenario(id='transitEncryptionEnabled_on_CMD_replication_group_causes_Advisory_condition')

    def test_scenario(self, scenario):
>       runner.run(scenario)

declarative_tests.py:92: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../declarative_test_fwk/runner.py:33: in run
    run_step(step)
../declarative_test_fwk/runner.py:82: in run_step
    assert_expectations(step)
../declarative_test_fwk/runner.py:125: in assert_expectations
    raise ae
../declarative_test_fwk/runner.py:122: in assert_expectations
    resource_helper.assert_expectations(step.verb, step.input_data, step.expectations, reference)
../declarative_test_fwk/helper.py:97: in assert_expectations
    self._assert_conditions(expectations, reference)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <e2e.declarative_tests.ReplicationGroupHelper object at 0x10627db50>
expectations = {'status': {'conditions': {'ACK.Advisory': 'True'}}}
reference = CustomResourceReference(group='elasticache.services.k8s.aws', version='v1alpha1', plural='replicationgroups', name='test-5sebujxj5bsa111bdv1touzi2ut52kl', namespace='default')

    def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceReference):
>       assert True == False
E       AssertionError

../declarative_test_fwk/helper.py:106: AssertionError
```

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,108 @@
...
+"""
+
+from declarative_test_fwk import helper
+
+
+class Step:
+    """
+    Represents a declarative test step
+    """
+    indent = "\t\t"
```
jaypipes, line 24: unused?

test/e2e/scenarios/scenario1.yaml
```diff
@@ -0,0 +1,40 @@
+id: "RG_CMD_basic_create_update"
```
jaypipes, line 1: recommend naming the file `test/e2e/scenarios/replication_group/cmd_basic_create_update.yaml` or something like that instead of `scenario1.yaml` :)

test/e2e/scenarios/scenario1.yaml
```diff
@@ -0,0 +1,40 @@
...
+description: "In this test we create RG with absolutely minimum configuration and try to update it."
+usecases:
+  - replication_group
+  - cmd
+#marks:
+#  - slow
+#  - blocked
```
jaypipes, line 8: What's the difference between usecases and marks?

test/e2e/scenarios/scenario1.yaml
```diff
@@ -0,0 +1,40 @@
...
+steps:
+  - id: "create_CMD_replication_group"
+    description: "Initial config"
+    create:
+      spec:
+        engine: redis
+        replicationGroupID: test$RANDOM_SUFFIX
+        replicationGroupDescription: Basic create and update of CMD RG
+        cacheNodeType: cache.t3.micro
+        numNodeGroups: 1
```
jaypipes, line 23: This is cool to allow `spec` values directly in the scenario YAML definition. I'd recommend also allowing to load from a filepath as wel.

test/e2e/scenarios/scenario1.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - replication_group
+  - cmd
+#marks:
+#  - slow
+#  - blocked
+customResourceReference:
+  apiVersion: $CRD_GROUP/$CRD_VERSION
+  kind: ReplicationGroup
+  metadata:
+    name: test$RANDOM_SUFFIX
```
jaypipes, line 13: where does `$RANDOM_SUFFIX` come from?

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+        :return: None
+        """
+        # condition assertion contains wait logic
+        self._assert_conditions(expectations, reference)
+
+        # conditions expectations met, now check current resource against expectations
+        resource = k8s.get_resource(reference)
+        self.assert_items(expectations.get("status"), resource.get("status"))
+
+        # self._assert_state(expectations.get("spec"), resource)  # uncomment to support spec assertions
```
jaypipes, line 100: This is kinda funky because you are reading the CR from the Kubernetes API server multiple times (hidden to the user) but the test scenario acts like what is being asserted is simply the returned CR from the Create or Patch operation.

I feel like a more robust solution, if slightly more verbose, would be to have an explicit `wait` part of the `Step` structure that occurs before the `expect` stanza.

So, instead of this:

```yaml
steps:
  - id: "create_CMD_replication_group"
    description: "Initial config"
    create:
      spec:
        engine: redis
        replicationGroupID: test$RANDOM_SUFFIX
        replicationGroupDescription: Basic create and update of CMD RG
        cacheNodeType: cache.t3.micro
        numNodeGroups: 1
      expect:
        status:
          conditions:
            ACK.ResourceSynced: "True"
          status: "available"
```

You would do something like this:

```yaml
steps:
  - id: "create_CMD_replication_group"
    description: "Initial config"
    create:
      spec:
        engine: redis
        replicationGroupID: test$RANDOM_SUFFIX
        replicationGroupDescription: Basic create and update of CMD RG
        cacheNodeType: cache.t3.micro
        numNodeGroups: 1
    wait:
      status:
         conditions:
           ACK.ResourceSynced:
             value: "True"
             timeout: 30
    # These assertions are made on the resulting CR from a Kuberetes Describe operation
    # after all waits have successfully completed
    expect:
      status: "available"
```

That way, it's more explicit that there are actually distinct reads to the Kubernetes API server, with one set of reads having a retry/wait semantic and the other not.

jaypipes (commented):
@kumargauravsharma once again, I'm diggin' the move to declarative testing. ++ on that!

I have some suggestions inline that I think might make the test scenarios easier to write and read.

test/e2e/scenarios/scenario1.yaml
kumargauravsharma: I added some details related to it at: https://github.com/aws-controllers-k8s/elasticache-controller/pull/40#issuecomment-881827614 

I think currently marks are used to include tests that are by default excluded otherwise.
The intent of usecases was to have 'selector'/'filter' behavior such that when it is specified to pytest run command,   only tests that match the supplied usecase are included in the test run.

Note: the behavior is not yet implemented.

test/declarative_test_fwk/model.py
kumargauravsharma: Removed the unused variable.

test/e2e/scenarios/scenario1.yaml
kumargauravsharma: Updated file name.

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,107 @@
...
+        return self.config.get("description", "")
+
+    def resource_kind(self) -> str:
+        return self.input_data.get("kind")
+
+    def __str__(self) -> str:
+        return f"Step(id='{self.id()}')"
+
+    def __repr__(self) -> str:
+        return str(self)
```
RedbackThomson, line 72: Doesn't have to be in this PR, but this is exactly what the `dataclass` decorator helps with. It also writes your `__init__` for you, in terms of providing default values and constructor parameters for instance variables.

test/declarative_test_fwk/runner.py
```diff
@@ -0,0 +1,142 @@
...
+import logging
+from acktest.k8s import resource as k8s
+
+
+def run(scenario: model.Scenario) -> None:
+    """
+    Runs steps in the scenario
+    """
+    if not scenario:
+        return
```
RedbackThomson, line 29: Not necessary, it's not an optional parameter?

test/declarative_test_fwk/runner.py
```diff
@@ -0,0 +1,142 @@
...
+"""
+
+from declarative_test_fwk import model, helper
+import pytest
+import sys
+import logging
+from acktest.k8s import resource as k8s
+
+
+def run(scenario: model.Scenario) -> None:
```
RedbackThomson, line 24: I think in keeping with the OOP style of `Scenario` and `Step`, there should also be a `DeclarativeTestRunner` class that manages and loads `scenario` instances.

test/declarative_test_fwk/runner.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Runs step
+    """
+    if not step:
+        return
+
+    if not step.verb:
+        logging.warning(
+            f"skipped: {step}. No matching verb found."
+            f" Supported verbs: create, patch, delete.")
```
RedbackThomson, line 72: This check can be done in the `Step` constructor, earlier?

test/e2e/scenarios/cmd_basic_create_update.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - replication_group
+  - cmd
+#marks:
+#  - slow
+#  - blocked
+customResourceReference:
+  apiVersion: $CRD_GROUP/$CRD_VERSION
+  kind: ReplicationGroup
+  metadata:
+    name: test$RANDOM_SUFFIX
```
RedbackThomson, line 13: I would strongly urge you to rename this to `namePrefix` instead of name and assume that every metadata `name` is suffixed. Any tests that run in parallel would otherwise try to create resources with the same name and break soak tests.

test/e2e/scenarios/cmd_basic_create_update.yaml
```diff
@@ -0,0 +1,40 @@
...
+steps:
+  - id: "create_CMD_replication_group"
+    description: "Initial config"
+    create:
+      spec:
```
RedbackThomson, line 18: +1 to Jay's comment about this being in a separate resource file.

test/declarative_test_fwk/runner.py
```diff
@@ -0,0 +1,142 @@
...
+        return
+
+    logging.info(f"Execute: {step}")
+    if step.verb == "create":
+        create_resource(step)
+    elif step.verb == "patch":
+        patch_resource(step)
+    elif step.verb == "delete":
+        delete_resource(step)
+    assert_expectations(step)
```
echen-98, line 82: As discussed above, would prefer an explicit wait in between performing the action and asserting state

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def resource_helper(resource_kind: str):
+    """
+    Decorator to discover Custom Resource Helper
+    :param resource_kind: custom resource kind
+    """
+    def registrar(cls):
+        TEST_HELPERS[resource_kind.lower()] = cls
+        logging.info(f"Registered ResourceHepler: {cls.__name__} for custom resource kind: {resource_kind}")
```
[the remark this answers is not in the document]
RedbackThomson: Oh actually the Python syntax for checking subclass is `issubclass(cls, ResourceHelper)`

test/e2e/scenarios/scenario1.yaml
kumargauravsharma: Updated code. `create`, `patch`, `delete` now support reading custom resource from filepath. 

Filepath is resolved by taking `resource_directory` parameter into account:
```
loader.load_scenario(scenario_file_path, resource_directory, input_replacements)
```

Example:
```
steps:
  - id: "create_CMD_replication_group"
    description: "Initial config"
    create: "replicationgroup_cmd.yaml"
```

test/e2e/scenarios/scenario1.yaml
kumargauravsharma: Method:
`load_scenario(scenario_file: Path, resource_directory: Path = None, replacements: dict = {}) -> model.Scenario` 

sets `RANDOM_SUFFIX` in `replacements` dictionary before loading the scenario file. 

Each scenario gets different random suffix.

test/declarative_test_fwk/helper.py
kumargauravsharma: Updated code.
**`wait` can be specified as:**

1. Wait for 5 minutes
```
    wait: 5
```
2. Wait on condition expected value - waits for default timeout (30 minutes)
```
    wait:
      status:
        conditions:
          ACK.Terminal: "True"
```
I kept the `wait` details open ended, so that support to wait on other properties of custom resource can be be added later.

3. Wait on specific condition details with optional `timeout`:
```
    wait:
      status:
        conditions:
          ACK.ResourceSynced:
            status: "True"
            timeout: 30     # timeout in minutes, if not specified defaults to 30 minutes
```

**Assert on condition `message`:**

Conditions can be specified with expected `message` (both in `wait` as well as `expect`):
```
    status:
        conditions:
          ACK.Terminal: "True"
          ACK.ResourceSynced:
            status: "False"
            message: "Expected message ..."

```

test/declarative_test_fwk/helper.py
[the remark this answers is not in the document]
kumargauravsharma: Update code.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+"""
+
+import logging
+from time import sleep
+from acktest.k8s import resource as k8s
+
+TEST_HELPERS = dict()
+
+
+def resource_helper(resource_kind: str):
```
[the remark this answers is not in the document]
kumargauravsharma: Renamed to `register_resource_helper`

test/e2e/tests/test_scenarios.py
```diff
@@ -0,0 +1,99 @@
...
+
+@helper.resource_helper("ReplicationGroup")
+class ReplicationGroupHelper(helper.ResourceHelper):
+    def assert_expectations(self, verb: str, input_data: dict, expectations: dict,
+                            reference: k8s.CustomResourceReference):
+        # default assertions
+        super().assert_expectations(verb, input_data, expectations, reference)
+
+        # perform custom server side checks based on:
+        # verb, input data to verb, expectations for given resource
```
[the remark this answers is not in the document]
kumargauravsharma: Removed the overridden method. 
I kept it in the earlier revision of the code to highlight the ability of custom resource helper (specialized in service controller codebase) to implement custom assertions - for example invoking aws cli/api for validations.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+        Creates custom resource
+        :param input_data: resource details
+        :param input_replacements: input replacements
+        :return: k8s.CustomResourceReference, created custom resource
+        """
+        reference = self.custom_resource_reference(input_data, input_replacements)
+        _ = k8s.create_custom_resource(reference, input_data)
+        resource = k8s.wait_resource_consumed_by_controller(reference, wait_periods=10)
+        assert resource is not None
+        return (reference, resource)
```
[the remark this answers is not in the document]
kumargauravsharma: Each step can have only one of the verbs.

test/declarative_test_fwk/runner.py
kumargauravsharma: Removed this check.

test/declarative_test_fwk/runner.py
kumargauravsharma: Runner is stateless.
To provide flexibility to service conroller test code and integration with pytest, current design provides `loader.py` and `runner.py` with stateless methods, while model provides objects to hold state (Scenario, Step).
Execution/Runtime state utilized pytest framework allowing parallel execution of scenarios that share custom resource helper instance which is also stateless but declared as class to help specializations.

test/declarative_test_fwk/runner.py
kumargauravsharma: With the latest code, this check was not necessary. Thus, removed it.

test/e2e/scenarios/cmd_basic_create_update.yaml
kumargauravsharma: `RANDOM_SUFFIX` is added to the `repacements` before loading the scenario.
Each scenario gets a different value for it, thus parallel runs do not cause same name.

test/e2e/scenarios/cmd_basic_create_update.yaml
kumargauravsharma: Added support for resource yaml file name as value.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class ResourceHelper:
+    """
+    Provides generic verb (create, patch, delete) methods for custom resources.
+    Keep its methods stateless. Methods are on instance to allow specialization.
+    """
+    DEFAULT_WAIT_SECS = 30
+
+    def create(self, input_data: dict, input_replacements: dict = {}):
```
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+        logging.debug(f"No-op wait_for_delete()")
+
+    def resource_plural(self, resource_kind: str) -> str:
+        """
+        Provide plural string for supplied custom resource kind
+        Override as needed
+        :param resource_kind: custom resource kind
+        :return: plural string
+        """
+        return resource_kind.lower() + "s"
```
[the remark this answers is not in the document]
kumargauravsharma: Updated code to use following method signature:
```
def register_resource_helper(resource_kind: str, resource_plural: str)
```

test/declarative_test_fwk/loader.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def load_scenario(scenario_file: Path, replacements: dict = {}) -> model.Scenario:
+    """
+    Loads scenario from given scenario_file
+    :param scenario_file: yaml file containing scenarios
+    :param replacements: input replacements
+    :return: Iterable scenarios
+    """
+    scenario_name = scenario_file.name.split(".yaml")[0]
```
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

test/declarative_test_fwk/loader.py
```diff
@@ -0,0 +1,82 @@
...
+def idfn(scenario_file_full_path: Path) -> str:
+    """
+    Provides scenario file name as scenario test id
+    :param scenario: test scenario file path
+    :return: scenario test id string
+    """
+    return scenario_file_full_path.name
+
+
+def marks(scenario: model.Scenario) -> list:
```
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,107 @@
...
+from declarative_test_fwk import helper
+
+
+class Step:
+    """
+    Represents a declarative test step
+    """
+
+    def __init__(self, config: dict, custom_resource_details: dict, replacements: dict = {}):
+        self.config = config
```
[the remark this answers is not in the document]
kumargauravsharma: Though service controller specific test file does not construct the `Scenario`, `Step` objects as these are created as result of `loader.load_scenario()`. I have updated code and added following to model.

**Related to `Step`**
```
class Step:
    def __init__(self, resource_directory: Path, config: StepDict, custom_resource_details: dict, replacements: dict = {}):

CreateDict = TypedDict("CreateDict", {
        "apiVersion": str,
        "kind": str,
        "metadata": Dict,
        "spec": Dict
    }, total=False)

PatchDict = TypedDict("PatchDict", {
        "apiVersion": str,
        "kind": str,
        "metadata": Dict,
        "spec": Dict
    }, total=False)

DeleteDict = TypedDict("DeleteDict", {
        "apiVersion": str,
        "kind": str,
        "metadata": Dict
    }, total=False)

ExpectDict = TypedDict("ExpectDict", {
        "spec": Dict,
        "status": Dict
    }, total=False)

StepDict = TypedDict("StepDict", {
        "id": str,
        "description": str,
        Verb.create.name: Union[str, CreateDict],
        Verb.patch.name: Union[str, PatchDict],
        Verb.delete.name: Union[str, DeleteDict],
        "wait": Union[int, Dict],
        "expect": ExpectDict
    }, total=False)
```

**Related to `Scenario`**
```
class Scenario:
    def __init__(self, resource_directory: Path, config: ScenarioDict, replacements: dict = {}):


ResourceReferenceDict = TypedDict("CreateDict", {
        "apiVersion": str,
        "kind": str,
        "metadata": Dict
    }, total=False)

ScenarioDict = TypedDict("ScenarioDict", {
        "id": str,
        "description": str,
        "marks": List[str],
        "usecases": List[str],
        "customResourceReference": ResourceReferenceDict,
        "steps": List[StepDict]
    }, total=False)
```

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,107 @@
...
+        self.replacements = replacements
+
+        self.verb = None
+        self.input_data = {}
+        self.expectations = None
+
+        # (k8s.CustomResourceReference, ko) to teardown
+        self.teardown_list = []
+
+        supported_verbs = ["create", "patch", "delete"]
```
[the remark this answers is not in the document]
kumargauravsharma: Added enum for verbs.

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,107 @@
...
+        for verb in supported_verbs:
+            if verb not in self.config:
+                continue
+            self.verb = verb
+            self.input_data = self.config.get(verb)
```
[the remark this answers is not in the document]
kumargauravsharma: One step can have only one verb in it and I wanted to check if it has been specified in the yaml.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # self._assert_state(expectations.get("spec"), resource)  # uncomment to support spec assertions
+
+    def _assert_conditions(self, expectations: dict, reference: k8s.CustomResourceReference):
+        expect_conditions: dict = {}
+        if "status" in expectations and "conditions" in expectations["status"]:
+            expect_conditions = expectations["status"]["conditions"]
+
+        for (condition_name, condition_value) in expect_conditions.items():
+            assert k8s.wait_on_condition(reference, condition_name, condition_value, wait_periods=30)
```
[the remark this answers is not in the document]
kumargauravsharma: Updated the code for wait logic. Scenario step provides details about `wait` before expectations. Expectation assertions are validated without wait.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Asserts state against supplied expectations
+        Override it as needed for custom verifications
+        :param expectations: dictionary with items to assert in state
+        :param state: dictionary with items
+        :return: None
+        """
+        if not expectations:
+            return
+        if not state:
```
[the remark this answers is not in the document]
kumargauravsharma: Added comments in the code.
`state` check will be performed when there are some expectations. When `state is None` the assertion fails and helps in debugging.

test/declarative_test_fwk/runner.py
kumargauravsharma: Updated code: added `wait` before `expectations` and expectations check does not wait.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+class ResourceHelper:
+    """
+    Provides generic verb (create, patch, delete) methods for custom resources.
+    Keep its methods stateless. Methods are on instance to allow specialization.
+    """
+    DEFAULT_WAIT_SECS = 30
+
+    def create(self, input_data: dict, input_replacements: dict = {}):
+        """
+        Creates custom resource
```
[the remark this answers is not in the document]
kumargauravsharma: Updated documentation.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,173 @@
...
+from time import sleep
+from acktest.k8s import resource as k8s
+
+TEST_HELPERS = dict()
+
+
+def resource_helper(resource_kind: str):
+    """
+    Decorator to discover Custom Resource Helper
+    :param resource_kind: custom resource kind
```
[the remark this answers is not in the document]
kumargauravsharma: Thanks for the suggestion. 
Updated PyDocs.

test/e2e/scenarios/cmd_basic_create_update.yaml
RedbackThomson: I understand that, but that requires developers adding `$RANDOM_SUFFIX` at the end of every `metadata.name`. I think we still need `$RANDOM_SUFFIX` for when developers want to create a random resource name (such as randomly name their cluster), but for `metadata.name` I think we should take control. Another suggestion is that we just don't allow adding `metadata` in these scenario files at all, and we use the ID of the scenario as the name? Further, it would be great to run each set of scenarios in a separate namespace.

test/e2e/tests/test_scenarios.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Session scoped fixture to bootstrap service resources and teardown
+    provides input replacements for test scenarios.
+    Eliminates the need for:
+     - bootstrap.yaml and
+     - call to <service_controller_test_dir>/service_bootstrap.py from test-infra
+     - call to <service_controller_test_dir>/service_cleanup.py from test-infra
+    """
+
+    resources_dict = service_bootstrap.service_bootstrap()
```
RedbackThomson, line 57: I do not support using `service_bootstrap` and `service_cleanup` inside fixtures. These resources are especially designed to last longer than the scope of a single test invocation. That is, you create them once, run several PyTest runs, and then clean them up. 

I would encourage you to use `get_bootstrap_resources` here, instead: https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/test/e2e/bootstrap_resources.py#L35

RedbackThomson (changes requested):
Getting close, now. Changes to the wait are really impressive! Inline comments about Python-esque typing and some general questions. Thanks!

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,251 @@
...
+                    actual_condition = k8s.get_resource_condition(reference, condition_name)
+                    assert actual_condition is not None
+                    assert expected_value == actual_condition.get("status")
+
+            elif type(expected_value) is dict:
+                # Example:
+                # ACK.ResourceSynced:
+                #     status: "False"
+                #     message: "Expected message ..."
+                #     timeout: 30 # minutes
```
jaypipes, line 162: Seconds, not minutes :)

test/declarative_test_fwk/helper.py
jaypipes: >     1. Wait for 5 minutes

Seconds, not minutes :)

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,180 @@
...
+                continue
+            self.verb = verb
+            self.input_data = self.config.get(verb.name)
+            if type(self.input_data) is str:
+                if self.input_data.endswith(".yaml"):
+                    # load input data from resource file
+                    resource_file_name = self.input_data
+                    resource_file_path = Path(join(resource_directory, resource_file_name))
+                    self.input_data = load_resource_file(
+                        resource_file_path.parent, resource_file_path.stem, additional_replacements=replacements)
```
jaypipes, line 102: ++

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,180 @@
...
+        self.input_data = {}
+        self.expectations = None
+
+        # (k8s.CustomResourceReference, ko) to teardown
+        self.teardown_list = []
+
+        for verb in list(Verb):
+            if verb.name not in self.config:
+                continue
+            self.verb = verb
```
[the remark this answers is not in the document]
jaypipes: The `break` on line 110 means the first matched verb will be used. That said, I agree with @RedbackThomson that it's non-obvious code here and potentially needs either a check *or* a code comment explaining this is on purpose :)

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,180 @@
...
+        "metadata": Dict
+    }, total=False)
+
+# fields in a test scenario
+ScenarioDict = TypedDict("ScenarioDict", {
+        "id": str,
+        "description": str,
+        "marks": List[str],
+        "usecases": List[str],
+        "customResourceReference": ResourceReferenceDict,
```
jaypipes, line 149: nit: I think "resource" is fine for the key here instead of the longer "customResourceReference". There's no information lost in the briefer name.

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,180 @@
...
+        return self.input_data.get("kind")
+
+    def __str__(self) -> str:
+        return f"Step(id='{self.id()}')"
+
+    def __repr__(self) -> str:
+        return str(self)
+
+# fields for 'customResourceReference' field in a test Scenario
+ResourceReferenceDict = TypedDict("CreateDict", {
```
jaypipes, line 137: maybe just call this `ResourceDict` (see below). Also, change `"CreateDict"` to `"ResourceDict"` :)

test/declarative_test_fwk/runner.py
```diff
@@ -0,0 +1,226 @@
...
+    teardown_failures = []
+    # tear down steps in reverse order
+    for step in reversed(scenario.steps()):
+        try:
+            teardown_step(step)
+        except:
+            error = f"Failed to teardown: {step}. " \
+                    f"Unexpected error: {sys.exc_info()[0]}"
+            teardown_failures.append(error)
+            logging.debug(error)
```
jaypipes, line 60: do we need this `logging.debug` line if we're using `logging.error` below on line 65?

test/declarative_test_fwk/runner.py
```diff
@@ -0,0 +1,226 @@
...
+    Returns:
+        None
+    """
+
+    resource_helper = helper.get_resource_helper(step.resource_kind())
+    if not reference:
+        logging.debug(f"delete:  {step}")
+        reference = resource_helper.custom_resource_reference(step.input_data, step.replacements)
+
+    resource_helper.delete(reference)
```
jaypipes, line 147: Do we need to remove the resource reference from our step's teardown_list so that double-deletes are not attempted if our test steps manually call `delete_resource`?

test/e2e/scenarios/cmd_basic_create_update.yaml
```diff
@@ -0,0 +1,48 @@
+id: "RG_CMD_basic_create_update"
+description: "In this test we create RG with absolutely minimum configuration and try to update it."
+usecases:
+  - replication_group
+  - cmd
```
jaypipes, line 5: I guess I still don't think it's necessary to have a separate "usecases" key instead of a generic "marks" key (like the one commented out below).

test/e2e/tests/test_scenarios.py
```diff
@@ -0,0 +1,94 @@
...
+    Helper for replication group scenarios.
+    Overrides methods as required for custom resources.
+    """
+
+    def wait_for_delete(self, reference: k8s.CustomResourceReference):
+        logging.debug(f"ReplicationGroupHelper - wait_for_delete()")
+        ec = boto3.client("elasticache")
+        waiter = ec.get_waiter('replication_group_deleted')
+        # throws exception if wait fails
+        waiter.wait(ReplicationGroupId=reference.name)
```
jaypipes, line 43: :+1:

jaypipes (changes requested):
This is looking excellent, @kumargauravsharma, thank you! Some very small questions and nits inline, but I'm good to merge this after addressing the one major issue @RedbackThomson brought up about not using `service_bootstrap` from a session-scoped pytest fixture.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,251 @@
...
+        resource_kind: custom resource kind
+        resource_plural: custom resource kind plural
+
+    Returns:
+        wrapper
+    """
+
+    def registrar(cls):
+        if issubclass(cls, ResourceHelper):
+            TEST_HELPERS[resource_kind.lower()] = cls
```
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

test/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,251 @@
...
+        resource = k8s.get_resource(reference)
+        if not resource:
+            logging.warning(f"ResourceReference {reference} not found. Not invoking k8s delete api.")
+            return
+
+        k8s.delete_custom_resource(reference)
+        sleep(self.DEFAULT_WAIT_SECS)
+        self.wait_for_delete(reference) # throws exception if wait fails
+
+    def assert_expectations(self, verb: str, input_data: dict, expectations: dict, reference: k8s.CustomResourceReference) -> None:
```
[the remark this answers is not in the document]
kumargauravsharma: Added ExpectDict

test/declarative_test_fwk/loader.py
```diff
@@ -0,0 +1,101 @@
...
+
+    Args:
+        scenarios_directory: directory containing scenarios yaml files
+
+    Returns:
+        Iterable scenarios for pytest parameterized fixture
+    """
+
+    scenarios_list = []
+    for scenario_file in os.listdir(scenarios_directory):
```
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

test/declarative_test_fwk/loader.py
```diff
@@ -0,0 +1,101 @@
...
+        resource_directory: Path to custom resources directory
+        replacements: input replacements
+
+    Returns:
+        Scenario reference
+    """
+
+    scenario_name = scenario_file.stem
+    replacements = replacements.copy()
+    replacements["RANDOM_SUFFIX"] = random_suffix_name("", 32)
```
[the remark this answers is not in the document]
kumargauravsharma: Changed the random suffix length to 16

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,180 @@
...
+class Verb(Enum):
+    """
+    Verb for custom resource in a test step.
+    """
+    create = auto()
+    patch = auto()
+    delete = auto()
+
+# fields for 'create' Verb in a test step
+CreateDict = TypedDict("CreateDict", {
```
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

test/declarative_test_fwk/model.py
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

test/declarative_test_fwk/model.py
```diff
@@ -0,0 +1,180 @@
...
+        self.expectations = self.config.get("expect")
+
+    def id(self) -> str:
+        return self.config.get("id", "")
+
+    def description(self) -> str:
+        return self.config.get("description", "")
+
+    def resource_kind(self) -> str:
+        return self.input_data.get("kind")
```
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

test/e2e/tests/test_scenarios.py
kumargauravsharma: Updated code and used `get_bootstrap_resources`

test/declarative_test_fwk/helper.py
kumargauravsharma: Updated code to consider the value in seconds.

test/declarative_test_fwk/model.py
kumargauravsharma: Renamed to "resource"

test/declarative_test_fwk/model.py
kumargauravsharma: Updated code and used "ResourceDict"

test/declarative_test_fwk/runner.py
kumargauravsharma: Removed debug statement.

test/declarative_test_fwk/runner.py
kumargauravsharma: As one step can only have one verb, the list is populated when verb is "create" and it is cleared during `runner.teardown_step()`

test/e2e/scenarios/cmd_basic_create_update.yaml
kumargauravsharma: Removed the `usecases`

kumargauravsharma:
Thanks @RedbackThomson, @jaypipes 
I have addressed the review comments and raised a new revision of the code.
Please review.

test/e2e/declarative_test_fwk/helper.py
```diff
@@ -0,0 +1,261 @@
...
+            input_replacements: input replacements
+
+        Returns:
+            k8s.CustomResourceReference, created custom resource
+        """
+
+        reference = self.custom_resource_reference(input_data, input_replacements)
+        _ = k8s.create_custom_resource(reference, input_data)
+        resource = k8s.wait_resource_consumed_by_controller(reference, wait_periods=10)
+        assert resource is not None
```
A-Hilaly, line 73: Is it OK to use `assert` statements in non unit test files? I'm not very experienced with python but i always thought that it's discouraged to use `assert` in libraries/functions that be imported by other projects.

A-Hilaly (commented):
Looks really good @kumargauravsharma ! I left one small question

test/e2e/scenarios/cmd_basic_create_update.yaml
RedbackThomson: Any comments on this @kumargauravsharma ? I still feel it leaves room for mistakes as it currently exists. If everyone is just going to have to append $RANDOM_SUFFIX anyway, then lets just do it for them?

RedbackThomson (approved):
Yeah this looks like it's solid. I still stand by my comment for the `namePrefix` part, but I'm happy with everything else.

test/e2e/declarative_test_fwk/helper.py
kumargauravsharma: As the file is used in a test context I used the assert so that it helps in debug if it fails.
As we use this framework in e2e tests, we can evolve the code to keep/remove it later.

kumargauravsharma:
@RedbackThomson 
For `$RANDOM_SUFFIX` I considered scenario where one resource refers the other resource by name in its Spec, then the current mechanism allows such configuration, which will be difficult to setup if framework adds the suffix automatically.


For the current PR, I think we can keep the current logic.
We can enhace the framework as we use it to automate test scenarios and if we find that auto appending the suffix makes it better.

nmvk (approved):
/lgtm

RedbackThomson:
@kumargauravsharma Alright. I'll take your word for that. Let's ship it

PR #42, opened 2021-07-16
regenerate code, fix replication group delete templates, upgrade to runtime 0.6.0

echen-98:
Description of changes:
Please check if the resulting `sdkDelete` is as intended.

### DIFF
--- a/pkg/resource/replication_group/sdk.go
+++ b/pkg/resource/replication_group/sdk.go
@@ -1381,29 +1381,29 @@ func (rm *resourceManager) newUpdateRequestPayload(
 func (rm *resourceManager) sdkDelete(
 	ctx context.Context,
 	r *resource,
-) (err error) {
+) (latest *resource, err error) {
 	rlog := ackrtlog.FromContext(ctx)
 	exit := rlog.Trace("rm.sdkDelete")
 	defer exit(err)
+	// if resource is already deleting, return requeue error; otherwise, initiate deletion
 	if isDeleting(r) {
-		return requeueWaitWhileDeleting
+		return r, requeueWaitWhileDeleting
 	}
 
 	input, err := rm.newDeleteRequestPayload(r)
 	if err != nil {
-		return err
+		return nil, err
 	}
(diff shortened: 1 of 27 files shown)


pkg/resource/replication_group/sdk.go
```diff
@@ -1381,29 +1381,28 @@ func (rm *resourceManager) newUpdateRequestPayload(
...
 	rm.metrics.RecordAPICall("DELETE", "DeleteReplicationGroup", err)
+	// delete call successful
 	if err == nil {
-		if foundResource, err := rm.sdkFind(ctx, r); err != ackerr.NotFound {
-			if isDeleting(foundResource) {
-				return requeueWaitWhileDeleting
-			}
-			return err
-		}
+		return r, requeueWaitWhileDeleting
```
kumargauravsharma, line 1403: This logic does not take values from `resp.ReplicationGroup`.
Please add a comment here about this behavior in the code.

It is expected to work fine, as the next reconciler run (after requeue wait interval) will fetch latest state of it and lines 1389-1390 above will ensure that the correct status is updated inside reconciler->patchResource.

pkg/resource/replication_group/testdata/test_suite.yaml
```diff
@@ -139,17 +138,15 @@ tests:
             - operation: DescribeEventsWithContext
               output_fixture: "events/read_many/rg_cmd_events.json"
         invoke: Delete
-        expect: # for the delete case we don't expect a new latest state or a non-nil error
-          latest_state: nil
+        expect:
```
kumargauravsharma, line 141: As there is no call to `sdkFind()` inside `sdkDelete()`, there is no need to mock `DescribeReplicationGroupsWithContext`, `ListAllowedNodeTypeModifications`  and `ListAllowedNodeTypeModifications` service APIs.

echen-98:
@kumargauravsharma addressed both comments

kumargauravsharma (approved):

PR #43, opened 2021-07-16
fix update of maintenance window

echen-98:
Issue #, if available:
Update of maintenance window was being "blocked" by custom code: multiple differences were triggering an update, and the custom update code considered the engine versions different even if `desired` had `6.x` and latest had `6.0.5`. 

Since this takes precedence over the generated code, the generated code was never reached (and the maintenance window was therefore never updated).

Description of changes:
- change `RemoveFromDelta` to be common to all resources
- remove function `engineVersionsDiffer` (+ tests) and replace it with calls to `delta.DifferentAt`
- handle case where only major version is specified in post-delta hook
- update `post_set_output` hook in `ReadOne` to retrieve the latest maintenance window
- e2e test for this scenario


pkg/common/delta.go
```diff
@@ -0,0 +1,42 @@
...
+			break
+		}
+	}
+
+	// if found, create a new slice and replace the original
+	if i != nil {
+		differences = append(differences[:*i], differences[*i+1:]...)
+		delta.Differences = differences
+	}
+}
```
kumargauravsharma, line 42: nit: This method may return boolean indicating if there was a removal.

pkg/resource/replication_group/delta_util.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/aws-controllers-k8s/elasticache-controller/pkg/common"
+)
+
+// remove non-meaningful differences from delta
+func filterDelta(
+	delta *ackcompare.Delta,
+	desired *resource,
+	latest *resource,
+) {
```
kumargauravsharma, line 30: Nice.

Not related to this PR but mentioning it here due to its context: I think this is where Server defaults can be handled: Remove differences from delta when corresponding spec field is nil if last applied spec also had that field nil.
* This will require access to "last applied spec" which needs to be kept somewhere for reference.

pkg/resource/replication_group/post_set_output.go
```diff
@@ -42,9 +42,13 @@ func (rm *resourceManager) updateSpecFields(
 	latestCacheCluster, err := rm.describeCacheCluster(ctx, resource)
 	if err == nil && latestCacheCluster != nil {
 		setEngineVersion(latestCacheCluster, resource)
+		setMaintenanceWindow(latestCacheCluster, resource)
 	}
 }
 
+//TODO: for all the fields here, reevaluate if the latest observed state should always be populated,
+// even if the corresponding field was not specified in desired
+
```
kumargauravsharma, line 51: I think it should be fine to populate observed state fields here, the filterDelta() should determine if the diff is to be ignored.

kumargauravsharma (approved):

kumargauravsharma:
It can be merged after successful e2e tests

pkg/resource/replication_group/post_set_output.go
echen-98: Agreed

pkg/resource/replication_group/delta_util.go
echen-98: Yep this should enable us to (at least temporarily) handle defaults before the framework-level solution is ready

pkg/common/delta.go
echen-98: ACK, not sure how we would use that return value but we can add it on later if there is a use for it

PR #48, opened 2021-08-06
Replication group: cluster mode disabled test for spec update

echen-98:
The purpose of this test is to evaluate whether the controller can take the API responses from the ElastiCache describe APIs called during `ReadOne` and map them to the latest object's Spec.

A reliable record of the object's latest state is necessary to taking the correct action in updating the resource (or leaving it as-is).


echen-98:
Closing this as we may not be taking this approach

PR #49, opened 2021-08-14
e2e tests to cover most replication group update scenarios

echen-98:
Description of changes:
This PR introduces additional e2e tests to cover most of the possible modifications to a replication group 

### DIFF
--- a/test/e2e/resources/replicationgroup_cme_ngc.yaml
+++ b/test/e2e/resources/replicationgroup_cme_ngc.yaml
@@ -1,0 +1,24 @@
+# CME replication group. "NGC" means that the field nodeGroupConfiguration is specified, with a detailed configuration
+# for each shard
+apiVersion: elasticache.services.k8s.aws/v1alpha1
+kind: ReplicationGroup
+metadata:
+  name: $RG_ID
+spec:
+  cacheNodeType: cache.t3.micro
+  engine: redis
+  nodeGroupConfiguration:
+    - nodeGroupID: $NGID1
+      primaryAvailabilityZone: us-west-2a
+      replicaAvailabilityZones:
+        - us-west-2b
+        - us-west-2c
+      replicaCount: 2
+    - nodeGroupID: $NGID2
+      primaryAvailabilityZone: us-west-2b
+      replicaAvailabilityZones:
+        - us-west-2c
+        - us-west-2a
+      replicaCount: 2
+  replicationGroupDescription: cluster-mode enabled RG
+  replicationGroupID: $RG_ID
--- a/test/e2e/resources/replicationgroup_fault_tolerance.yaml
+++ b/test/e2e/resources/replicationgroup_fault_tolerance.yaml
@@ -1,0 +1,14 @@
+# CMD replication group for testing modification fault tolerance features (multi AZ, auto failover, primary promotion)
+apiVersion: elasticache.services.k8s.aws/v1alpha1
+kind: ReplicationGroup
+metadata:
+  name: $RG_ID
+spec:
+  automaticFailoverEnabled: $AF_ENABLED
+  cacheNodeType: cache.t3.micro
+  engine: redis
+  multiAZEnabled: $MAZ_ENABLED
+  numNodeGroups: 1
+  replicasPerNodeGroup: 1
+  replicationGroupDescription: description
+  replicationGroupID: $RG_ID
--- a/test/e2e/resources/replicationgroup_rpng.yaml
+++ b/test/e2e/resources/replicationgroup_rpng.yaml
@@ -1,0 +1,15 @@
+# "RPNG" means that the field replicasPerNodeGroup is set (instead of nodeGroupConfiguration),
+# meaning that this replication group has a uniform configuration across all shards. This is one of the more
+# basic configurations as very few fields are specified and the resulting RG can either be CME or CMD
+# (depending on the number of specified node groups).
+apiVersion: elasticache.services.k8s.aws/v1alpha1
+kind: ReplicationGroup
+metadata:
+  name: $RG_ID
+spec:
+  cacheNodeType: cache.t3.micro
+  engine: redis
+  numNodeGroups: $NUM_NODE_GROUPS
+  replicasPerNodeGroup: $REPLICAS_PER_NODE_GROUP
+  replicationGroupDescription: cluster-mode enabled RG
+  replicationGroupID: $RG_ID
--- a/test/e2e/service_cleanup.py
+++ b/test/e2e/service_cleanup.py
@@ -86,19 +86,34 @@ def service_cleanup(config: dict):
     )
 
     try:
-        delete_sns_topic(resources.SnsTopicARN)
+        delete_sns_topic(resources.SnsTopic1)
     except:
-        logging.exception(f"Unable to delete SNS topic {resources.SnsTopicARN}")
+        logging.exception(f"Unable to delete SNS topic {resources.SnsTopic1}")
 
     try:
-        delete_security_group(resources.SecurityGroupID)
+        delete_sns_topic(resources.SnsTopic2)
     except:
-        logging.exception(f"Unable to delete VPC Security Group {resources.SecurityGroupID}")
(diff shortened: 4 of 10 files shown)


test/e2e/resources/replicationgroup_cme_ngc.yaml
```diff
@@ -0,0 +1,24 @@
...
+      primaryAvailabilityZone: us-west-2a
+      replicaAvailabilityZones:
+        - us-west-2b
+        - us-west-2c
+      replicaCount: 2
+    - nodeGroupID: $NGID2
+      primaryAvailabilityZone: us-west-2b
+      replicaAvailabilityZones:
+        - us-west-2c
+        - us-west-2a
```
jaypipes, line 21: Just for my own information, if `replicaCount` is `2`, and the `replicaAvailabilityZones` contains two AZs that are *not* the `primaryAvailabilityZone`, does that mean that there will be a total of 2 instances (the primary in the primary AZ and one replica in one of the replica AZs)? Or does that mean there will be a total of 3 instances (the primary in the primary AZ and one replica in each of the replica AZs)?

test/e2e/resources/replicationgroup_fault_tolerance.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: ReplicationGroup
+metadata:
+  name: $RG_ID
+spec:
+  automaticFailoverEnabled: $AF_ENABLED
+  cacheNodeType: cache.t3.micro
+  engine: redis
+  multiAZEnabled: $MAZ_ENABLED
+  numNodeGroups: 1
+  replicasPerNodeGroup: 1
```
jaypipes, line 12: does this include the primary?

test/e2e/resources/replicationgroup_rpng.yaml
```diff
@@ -0,0 +1,15 @@
+# "RPNG" means that the field replicasPerNodeGroup is set (instead of nodeGroupConfiguration),
+# meaning that this replication group has a uniform configuration across all shards. This is one of the more
+# basic configurations as very few fields are specified and the resulting RG can either be CME or CMD
+# (depending on the number of specified node groups).
```
jaypipes, line 4: Interesting. So CME or CMD basically just means whether there is more than one node group?

If this is the case, if I change an RG from 2 node groups to 1, that means the RG goes from cluster mode enabled to cluster mode disabled automagically?

test/e2e/tests/test_replicationgroup.py
```diff
@@ -140,18 +140,52 @@ def rg_cmd_fromsnapshot(bootstrap_resources, make_rg_name, make_replication_grou
 
 
 @pytest.fixture(scope="module")
-def rg_cmd_update_input(make_rg_name):
+def rg_cme_uneven_shards_input(make_rg_name):
     return {
-        "RG_ID": make_rg_name("rg-cmd-update"),
+        "RG_ID": make_rg_name("rg-cme-uneven-shards"),
+        "NGID1": '"1111"',
+        "NGID2": '"2222"'
```
jaypipes, line 147: instead of all the double-quote trimming, why not just name these `ng-1` and `ng-2`?

test/e2e/resources/replicationgroup_cme_ngc.yaml
echen-98: This should produce a total of 3 instances in the node group – 1 primary and 2 replicas

test/e2e/resources/replicationgroup_fault_tolerance.yaml
echen-98: This does not – there should always be one primary for every node group

test/e2e/resources/replicationgroup_rpng.yaml
echen-98: It's a little complicated. "CME" just means the RG has the ability to have more than 1 node group. It's possible to create a CME RG with 1 node group, by specifying a CME parameter group when creating the replication group (so many "groups"!). To your second question, it's also possible to scale in a CME RG from 2 node groups to 1 – but it will not change from CME to CMD.

In contrast, a CMD RG will always have 1 node group. I'm not aware if there is a way to toggle cluster mode on a RG after its creation.

I think this page should have more details about what's going on under the hood: https://redis.io/topics/cluster-tutorial

test/e2e/tests/test_replicationgroup.py
echen-98: From what I tried last time, I think there's some weird constraint on this field that it has to be a string of numeric characters (although I'd have to double check that).

When these inputs are injected into the appropriate `.yaml`, the outer quotes are stripped and the inner double quotes are left to clarify that the value is a string instead of an integer.

test/e2e/service_cleanup.py
```diff
@@ -86,19 +86,22 @@ def service_cleanup(config: dict):
     )
 
     try:
-        delete_sns_topic(resources.SnsTopicARN)
+        delete_sns_topic(resources.SnsTopic1)
+        delete_sns_topic(resources.SnsTopic2)
     except:
-        logging.exception(f"Unable to delete SNS topic {resources.SnsTopicARN}")
+        logging.exception(f"Unable to delete SNS topics {resources.SnsTopic1}, {resources.SnsTopic2}")
```
kumargauravsharma, line 92: Please have `try: except:` block for each individual `delete_sns_topic` call.
This will ensure that `SnsTopic2` delele is tries even if `SnsTopic1` delete gets exception.

test/e2e/service_cleanup.py
```diff
@@ -86,19 +86,22 @@ def service_cleanup(config: dict):
...
-        logging.exception(f"Unable to delete VPC Security Group {resources.SecurityGroupID}")
+        logging.exception(f"Unable to delete VPC Security Groups {resources.SecurityGroup1}, {resources.SecurityGroup2}")
 
     try:
-        delete_user_group(resources.UserGroupID)
+        delete_user_group(resources.UserGroup1)
+        delete_user_group(resources.UserGroup2)
     except:
-        logging.exception(f"Unable to delete ElastiCache User Group {resources.UserGroupID}")
+        logging.exception(f"Unable to delete ElastiCache User Groups {resources.UserGroup1}, {resources.UserGroup2}")
```
kumargauravsharma, line 104: Please have `try: except:` block for each individual `delete_user_group` call.

test/e2e/service_cleanup.py
```diff
@@ -86,19 +86,22 @@ def service_cleanup(config: dict):
...
-        logging.exception(f"Unable to delete SNS topic {resources.SnsTopicARN}")
+        logging.exception(f"Unable to delete SNS topics {resources.SnsTopic1}, {resources.SnsTopic2}")
 
     try:
-        delete_security_group(resources.SecurityGroupID)
+        delete_security_group(resources.SecurityGroup1)
+        delete_security_group(resources.SecurityGroup2)
     except:
-        logging.exception(f"Unable to delete VPC Security Group {resources.SecurityGroupID}")
+        logging.exception(f"Unable to delete VPC Security Groups {resources.SecurityGroup1}, {resources.SecurityGroup2}")
```
kumargauravsharma, line 98: Please have `try: except:` block for each individual `delete_security_group` call.

test/e2e/service_cleanup.py
```diff
@@ -116,9 +119,10 @@ def service_cleanup(config: dict):
         logging.exception(f"Unable to delete user {resources.NonDefaultUser}")
 
     try:
-        delete_log_group(resources.CWLogGroup)
+        delete_log_group(resources.CWLogGroup1)
+        delete_log_group(resources.CWLogGroup2)
     except:
-        logging.exception(f"Unable to delete CW log group {resources.CWLogGroup}")
+        logging.exception(f"Unable to delete CW log groups {resources.CWLogGroup1}, {resources.CWLogGroup2}")
```
kumargauravsharma, line 125: Please have `try: except:` block for each individual `delete_log_group` call.

kumargauravsharma (changes requested):

kumargauravsharma (approved):

PR #53, opened 2021-09-09
support Tags field for all resources (on create)

jaypipes:
Re-adds support for the Tags field on all resources. Support only
includes tag on create operations due to the separate Tagris API calls
that need to be made when updating a resource's tags.

Signed-off-by: Jay Pipes <jaypipes@gmail.com>

By submitting this pull request, I confirm that my contribution is made
under the terms of the Apache 2.0 license.


surajkota:
Few things I would recommend,
- take a decision on what to do about delta in tags. describe API for a resource in sagemaker does not return tags, so a.spec.Tags will always be equal to b.spec.Tags even if tags didn't actually change on the service side. (since we are doing a `resource.ko.deepCopy()` in sdkFind)
  - in sagemaker we ignored the tags for comparison since it is not useful anyway
  
- since the describe API didn't return tags, we added an assert in the integration test to verify tags actually get created.

kumargauravsharma (approved):

PR #56, opened 2021-09-11
standardize setting conditions on RGs

jaypipes:
Instead of returning a RequeueOnError when a replication group is in
deleting or modifying status, we now return the latest observed resource
after setting the Synced condition to ConditionFalse and a `nil` error.
This triggers a requeue of the resource in the primary reconciliation
loop, since `resourceReconciler.handleRequeues` automatically requeues
any resources that have a Synced condition set to ConditionFalse.

Removes some of the custom code that was setting a terminal condition
when the replication group was in `create-failed` state and replaced
with hook code that calls the ACK runtime's `pkg/condition.SetTerminal`
function.

DO NOT MERGE. Code not yet working due to issue 950.

Issue aws-controllers-k8s/community#948
Issue aws-controllers-k8s/community#949
Issue aws-controllers-k8s/community#950

By submitting this pull request, I confirm that my contribution is made
under the terms of the Apache 2.0 license.


kumargauravsharma (approved):
Approving this PR as further fixes are part of followup PR: https://github.com/aws-controllers-k8s/elasticache-controller/pull/57 where e2e tests passed.

PR #58, opened 2021-09-16
Add CacheClusterNotFound terminal code

echen-98:
This fixes the failing negative scenario where we attempt to create a replication group with a `primaryClusterID` of a node that doesn't exist. Removed the skip marker from the test as it passes locally.


echen-98:
From Prow run: ` [gw2] [ 45%] PASSED tests/test_replicationgroup.py::TestReplicationGroup::test_rg_invalid_primary `

kumargauravsharma (approved):
Thanks @echen-98

PR #59, opened 2021-09-22
fixes to log delivery modification

echen-98:
### Problem 1: log delivery status is not in sync with replication group status

This was fixed by moving the population of `Status.LogDeliveryConfigurations` to execute after each of Create, Update, and Describe (instead of just Describe). See changes in `custom_set_output.go`.

### Problem 2: stale state from modify API merged into `desired` upon spec patching

This was causing the desired and latest state to infinitely swap. Solution: re-copy `LogDeliveryConfigurations` from `desired.Spec` to `latest.Spec` after the modify API response is merged in to ensure spec patching has no effect on this field. See changes in `custom_set_output.go`.

 https://github.com/aws-controllers-k8s/code-generator/pull/201 was an alternate solution but we were recommended to avoid doing something like this.

### Problem 3: Controller attempts to act on diff when log delivery is disabled

Solution: added in a few lines in `filterDelta` to remove the diff if log delivery is disabled and the API response is `nil`.

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2021-10-12T16:29:01Z"
-  build_hash: 4b30ff5578e2f570d1c5b1741f3098be0d78e246
-  go_version: go1.16.5
+  build_date: "2021-10-14T18:44:40Z"
+  build_hash: 385779a205bea50e8762b76bc75cab957cf723b9
+  go_version: go1.15.2
   version: v0.15.1
-api_directory_checksum: 69c917d1e1ee2b196c23799a02417b920297bc74
+api_directory_checksum: 8e4808b17b3a814f0a624eee8d68a0b45ba5e2c4
 api_version: v1alpha1
 aws_sdk_go_version: v1.38.52
 generator_config_info:
-  file_checksum: 261e03a6b9fb25805b045293947f13e5674d2924
+  file_checksum: 28e3cd3119df8028a65eb0729a462b4354d27fa9
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/pkg/resource/replication_group/custom_set_output.go
+++ b/pkg/resource/replication_group/custom_set_output.go
@@ -15,6 +15,8 @@ package replication_group
 
 import (
 	"context"
+	"encoding/json"
+
 	svcapitypes "github.com/aws-controllers-k8s/elasticache-controller/apis/v1alpha1"
 	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	"github.com/aws/aws-sdk-go/service/elasticache"
@@ -54,6 +56,7 @@ func (rm *resourceManager) CustomCreateReplicationGroupSetOutput(
 	ko *svcapitypes.ReplicationGroup,
 ) (*svcapitypes.ReplicationGroup, error) {
 	rm.customSetOutput(resp.ReplicationGroup, ko)
+	rm.setAnnotationsFields(r, ko)
 	return ko, nil
 }
 
@@ -64,6 +67,16 @@ func (rm *resourceManager) CustomModifyReplicationGroupSetOutput(
 	ko *svcapitypes.ReplicationGroup,
 ) (*svcapitypes.ReplicationGroup, error) {
 	rm.customSetOutput(resp.ReplicationGroup, ko)
+
+	// reset latest.spec.LDC to original value in desired to prevent stale data
+	// from the modify API being merged back into desired upon spec patching
+	var logDeliveryConfig []*svcapitypes.LogDeliveryConfigurationRequest
+	for _, ldc := range r.ko.Spec.LogDeliveryConfigurations {
+		logDeliveryConfig = append(logDeliveryConfig, ldc.DeepCopy())
+	}
+	ko.Spec.LogDeliveryConfigurations = logDeliveryConfig
+
+	rm.setAnnotationsFields(r, ko)
 	return ko, nil
 }
 
@@ -142,31 +155,8 @@ func (rm *resourceManager) customSetOutput(
 		ko.Status.AllowedScaleDownModifications = nil
 		ko.Status.AllowedScaleUpModifications = nil
 	}
-}
 
-// newListAllowedNodeTypeModificationsPayLoad returns an SDK-specific struct for the HTTP request
-// payload of the ListAllowedNodeTypeModifications API call.
-func (rm *resourceManager) newListAllowedNodeTypeModificationsPayLoad(respRG *elasticache.ReplicationGroup) (
-	*svcsdk.ListAllowedNodeTypeModificationsInput, error) {
-	res := &svcsdk.ListAllowedNodeTypeModificationsInput{}
-
-	if respRG.ReplicationGroupId != nil {
-		res.SetReplicationGroupId(*respRG.ReplicationGroupId)
-	}
-
-	return res, nil
-}
-
-func (rm *resourceManager) customSetOutputSupplementAPIs(
-	ctx context.Context,
-	r *resource,
-	respRG *elasticache.ReplicationGroup,
-	ko *svcapitypes.ReplicationGroup,
-) error {
-	events, err := rm.provideEvents(ctx, r.ko.Spec.ReplicationGroupID, 20)
-	if err != nil {
-		return err
-	}
+	// populate status logDeliveryConfigurations struct
 	if respRG.LogDeliveryConfigurations != nil {
 		var f11 []*svcapitypes.LogDeliveryConfiguration
 		for _, f11iter := range respRG.LogDeliveryConfigurations {
@@ -210,6 +200,31 @@ func (rm *resourceManager) customSetOutputSupplementAPIs(
 	} else {
 		ko.Status.LogDeliveryConfigurations = nil
 	}
+}
+
+// newListAllowedNodeTypeModificationsPayLoad returns an SDK-specific struct for the HTTP request
+// payload of the ListAllowedNodeTypeModifications API call.
+func (rm *resourceManager) newListAllowedNodeTypeModificationsPayLoad(respRG *elasticache.ReplicationGroup) (
+	*svcsdk.ListAllowedNodeTypeModificationsInput, error) {
+	res := &svcsdk.ListAllowedNodeTypeModificationsInput{}
+
+	if respRG.ReplicationGroupId != nil {
+		res.SetReplicationGroupId(*respRG.ReplicationGroupId)
+	}
+
+	return res, nil
+}
+
+func (rm *resourceManager) customSetOutputSupplementAPIs(
+	ctx context.Context,
+	r *resource,
+	respRG *elasticache.ReplicationGroup,
+	ko *svcapitypes.ReplicationGroup,
+) error {
+	events, err := rm.provideEvents(ctx, r.ko.Spec.ReplicationGroupID, 20)
+	if err != nil {
+		return err
+	}
 	ko.Status.Events = events
 	return nil
 }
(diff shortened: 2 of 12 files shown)


echen-98:
Fixed an additional problem: needed to populate `enabled` for `latest.spec.logDeliveryConfigurations`, since the API response does not return `enabled` directly and this was causing an unnecessary delta in some cases.

Also added a step to `service_bootstrap` to clear out the Cloudwatch policy as discussed offline.

echen-98:
` [gw12] [ 70%] PASSED tests/test_replicationgroup.py::TestReplicationGroup::test_rg_log_delivery `

Adding a skip for any regressions. Will deal with those in a separate PR as discussed offline.

echen-98:
New approach: save the last requested configuration. This saves us from having to map `Status.LogDeliveryConfigurations.Status` back to `Spec.LogDeliveryConfigurations.Enabled`, and also from removing the delta when `desired` has `Spec.LogDeliveryConfigurations.Enabled` false but `latest` has `Spec.LogDeliveryConfigurations` nil. 

Further, this prevents a loop of modifications when an invalid configuration is provided by the user.

The new field, `Status.LastRequestedLogDeliveryConfigurations`, is populated only after successful create and modify calls. `Spec.LogDeliveryConfigurations` has been removed from generated delta comparison. Instead, we compare `desired.Spec.LogDeliveryConfigurations` to `latest.Status.LastRequestedLogDeliveryConfigurations` (see `delta_util.go`).

Fixed merge conflicts and awaiting results of test run.

echen-98:
` [gw12] [ 80%] PASSED tests/test_replicationgroup.py::TestReplicationGroup::test_rg_log_delivery `

jaypipes (commented):
Any reason not to use annotations for this kind of "sausage-making" internal state-tracking information? Adding actual fields to the `Status` for tracking the "last requested log delivery configurations" just seems like it's trying to emulate the behaviour of the AWS Console instead of aligning with the "just focus on setting the desired state of your resource and let the controller take care of actuating that" design that Kubernetes encourages.

pkg/resource/replication_group/delta_util.go
```diff
@@ -62,3 +79,74 @@ func engineVersionsMatch(
...
+// whether to invoke an update or not. This comes down to only trying to apply a particular configuration
+// once.
+func slowLogRequiresUpdate(desired *resource, latest *resource) bool {
+	var desiredConfig *svcapitypes.LogDeliveryConfigurationRequest
+	var lastRequestedConfig *svcapitypes.LogDeliveryConfiguration
+
+	// retrieve desiredConfig from desired Spec
+	if desired.ko.Spec.LogDeliveryConfigurations != nil {
+		for _, config := range desired.ko.Spec.LogDeliveryConfigurations {
+			if config.LogType != nil && *config.LogType == "slow-log" {
```
A-Hilaly, line 93: Instead of hard coding `"slow-log"` You can use the generate enum variables here https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/apis/v1alpha1/enums.go#L87-L91

pkg/resource/replication_group/custom_set_output.go
```diff
@@ -249,3 +262,51 @@ func (rm *resourceManager) provideEvents(
...
+func (rm *resourceManager) setLastRequestedLogDeliveryConfigurations(
+	r *resource,
+	ko *svcapitypes.ReplicationGroup,
+) {
+	var result []*svcapitypes.LogDeliveryConfiguration
+	if r.ko.Spec.LogDeliveryConfigurations != nil {
+		for _, specConfig := range r.ko.Spec.LogDeliveryConfigurations {
+			statusConfig := &svcapitypes.LogDeliveryConfiguration{}
+
+			statusConfig.DestinationDetails = specConfig.DestinationDetails.DeepCopy()
```
A-Hilaly, line 280: These have already been `DeepCopy`-ed in line 73.

pkg/resource/replication_group/delta_util.go
```diff
@@ -62,3 +79,74 @@ func engineVersionsMatch(
...
+		}
+	}
+
+	// if both nil, no update to be done.
+	// if desired nil and latest non-nil, then the intention could be to disable; update required.
+	// similarly, if desired non-nil and latest nil, then log delivery must be enabled; update required.
+	// if both non-nil, compare values to determine whether an update is required
+	if desiredConfig == nil && lastRequestedConfig == nil {
+		return false
+	} else if desiredConfig == nil || lastRequestedConfig == nil {
```
A-Hilaly, line 114: What if `latestRequestConfig != nil` && `desiredConfig == nil`? The if statement in line 114 will not catch it

pkg/resource/replication_group/delta_util.go
```diff
@@ -62,3 +79,74 @@ func engineVersionsMatch(
...
+		if ackcompare.HasNilDifference(desiredConfig.DestinationType, lastRequestedConfig.DestinationType) {
+			return true
+		} else if desiredConfig.DestinationType != nil && lastRequestedConfig.DestinationType != nil {
+			if *desiredConfig.DestinationType != *lastRequestedConfig.DestinationType {
+				return true
+			}
+		}
+
+		// types are different here but we can still compare nil-ness
+		if ackcompare.HasNilDifference(desiredConfig.Enabled, lastRequestedConfig.Status) {
```
A-Hilaly, line 130: I think you're missing a case here: If `desiredConfig.Enabled == nil` && `latestRequestConfig == &"false"`.
`ackcompare.HasNilDifference` will return true, but logically speaking they are equal and the function should return `false` instead

echen-98: Noted. I think the new approach with storing the last requested configuration in `Annotations` should solve this.

The current version is a bit of a hack since I am forcing the `LogDeliveryConfigurationsRequest` type into the `LogDeliveryConfigurations` type (since I could not introduce a `LogDeliveryConfigurationsRequest` type into `Status`, as it does not appear in an API response).

apis/v1alpha1/ack-generate-metadata.yaml
```diff
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2021-09-23T18:08:08Z"
-  build_hash: fe3ca1f5ee52a0e49be0f8458b4e5058a8f0fd47
-  go_version: go1.15
-  version: v0.14.1
-api_directory_checksum: 69c917d1e1ee2b196c23799a02417b920297bc74
+  build_date: "2021-09-30T01:02:22Z"
+  build_hash: 386373b002204b5a215388c4d91af91fc9372398
+  go_version: go1.15.2
+  version: v0.14.0
```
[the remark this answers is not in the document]
echen-98: Not intentional! Fixed it, and will switch to using annotations to store this state.

pkg/resource/replication_group/custom_set_output.go
```diff
@@ -249,3 +264,31 @@ func (rm *resourceManager) provideEvents(
...
+// to be called upon a successful creation or modification (but not read/describe),
+// setLastRequestedLogDeliveryConfigurations copies desired.Spec.LogDeliveryConfigurations
+// into the annotations of the object.
+// r represents desired, ko represents latest
+func (rm *resourceManager) setLastRequestedLogDeliveryConfigurations(
+	r *resource,
+	ko *svcapitypes.ReplicationGroup,
+) {
+	//TODO: if in the future we need to set other annotations fields, ensure all get persisted
+	// as opposed to each making a copy of desired and only modifying one field
```
RedbackThomson, line 277: You could modify this method signature to accept the annotations map, and modify only this key. Then in the parent functions, it from object meta and persist it back afterwards?

pkg/resource/replication_group/custom_set_output.go
```diff
@@ -249,3 +264,31 @@ func (rm *resourceManager) provideEvents(
 	}
 	return events, nil
 }
+
+// to be called upon a successful creation or modification (but not read/describe),
+// setLastRequestedLogDeliveryConfigurations copies desired.Spec.LogDeliveryConfigurations
+// into the annotations of the object.
+// r represents desired, ko represents latest
+func (rm *resourceManager) setLastRequestedLogDeliveryConfigurations(
```
RedbackThomson, line 272: Your comments are out of order?

pkg/resource/replication_group/custom_set_output.go
```diff
@@ -249,3 +264,31 @@ func (rm *resourceManager) provideEvents(
...
+	// copy desired annotations map into new map
+	desiredAnnotations := r.ko.ObjectMeta.GetAnnotations()
+	annotations := make(map[string]string)
+	for k, v := range desiredAnnotations {
+		annotations[k] = v
+	}
+
+	lastRequestedConfigs, err := json.Marshal(r.ko.Spec.LogDeliveryConfigurations)
+	if err != nil {
+		annotations[AnnotationLastRequestedLDCs] = "null"
```
RedbackThomson, line 288: Is null unmarshaled and handled properly in this case? Also, if this is null, will it break the functionality of the controller?

pkg/resource/replication_group/delta_util.go
```diff
@@ -37,6 +40,21 @@ func filterDelta(
 		}
 		// TODO: handle the case of a nil difference (especially when desired EV is nil)
 	}
+
+	// if server has given PreferredMaintenanceWindow a default value, no action needs to be taken
+	if delta.DifferentAt("Spec.PreferredMaintenanceWindow") {
+		if desired.ko.Spec.PreferredMaintenanceWindow == nil && latest.ko.Spec.PreferredMaintenanceWindow != nil {
+			common.RemoveFromDelta(delta, "Spec.PreferredMaintenanceWindow")
+		}
+	}
```
RedbackThomson, line 49: Rather than removing from delta after the fact, for default values, I've used the pattern of adding the default value to the object before the delta. See https://github.com/aws-controllers-k8s/s3-controller/blob/f4df629190ec0de88dfe92e22256490ce0775d06/pkg/resource/bucket/hook.go#L361-L372

test/e2e/tests/test_replicationgroup.py
```diff
@@ -640,6 +640,7 @@ def test_rg_associate_resources(self, rg_associate_resources_input, rg_associate
         # assert new state
         assert_associated_resources(rg_associate_resources_input['RG_ID'], sg_list, sns_topic, ug_list)
 
+    @pytest.mark.blocked  # TODO: remove when passing
```
RedbackThomson, line 643: No longer passing?

RedbackThomson (commented):
Yeah, a much cleaner implementation with the annotations. I have some inline comments. Nothing major, though.

test/e2e/tests/test_replicationgroup.py
echen-98: Yep, there were many fields using the always-present diff from the `preferredMaintenanceWindow` field as a crutch to invoke the modify API. I fixed that above which means this test is now failing, but I'll resolve that as a follow-up to this PR.

Thanks for the detailed comments/suggestions, I'll be responding to the rest in a bit!

pkg/resource/replication_group/custom_set_output.go
echen-98: Thanks for the suggestion! The latest version now uses this model

nmvk (approved):
LGTM

pkg/resource/replication_group/delta_util.go
```diff
@@ -62,3 +80,19 @@ func engineVersionsMatch(
...
+func logDeliveryRequiresUpdate(desired *resource) bool {
+	var desiredConfigs []*svcapitypes.LogDeliveryConfigurationRequest
+	var lastRequestedConfigs []*svcapitypes.LogDeliveryConfigurationRequest
+
+	desiredConfigs = desired.ko.Spec.LogDeliveryConfigurations
+
+	annotations := desired.ko.ObjectMeta.GetAnnotations()
+	if val, ok := annotations[AnnotationLastRequestedLDCs]; ok {
+		_ = json.Unmarshal([]byte(val), &lastRequestedConfigs)
+	}
```
A-Hilaly, line 95: error can ignored here for a reason?

echen-98: During marshaling, if there is an error, we set the value to the string literal "null" (as discussed above).

Whatever value we unmarshal from `annotations[AnnotationLastRequestedLDCs]` will either be "null" or some valid string representation of the `Spec.LogDeliveryConfigurations` field.

So it's hard to imagine a situation in which `Unmarshal` would result in a error. Even if it does and `lastRequestedConfigs` is populated incorrectly/incompletely, this would just result in adding a delta for `LogDeliveryConfigurations` and the controller would attempt another modification with the desired state, which I think is an okay outcome.

I considered adding logging here in the case of an error, but that would involve adding the resourceManager as a receiver on this function and consequently `filterDelta` and `newResourceDelta`, which may involve a change to the template for `delta.go` (unless there's an easier way to add logging that I'm not aware of).

Another possible solution is just to `panic()` here to convey the error, but that seems a little heavy-handed to me. 

I think the status quo is okay but let me know what you think.

PR #61, opened 2021-09-24
Enable Scale-In without NodeGroupConfiguration

nmvk:
Scale-In requires the node group to remove or node group to retain.

Updating logic to remove last node group if NodeGroupConfiguration isn't specified.

### DIFF
--- a/pkg/resource/replication_group/custom_update_api.go
+++ b/pkg/resource/replication_group/custom_update_api.go
@@ -472,6 +472,20 @@ func (rm *resourceManager) newUpdateShardConfigurationRequestPayload(
 		res.SetNodeGroupCount(*desiredShardsCount)
 	}
 
+	// If desired nodegroup count (number of shards):
+	// - increases, then (optional) provide ReshardingConfiguration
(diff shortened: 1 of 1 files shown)


pkg/resource/replication_group/custom_update_api.go
```diff
@@ -515,6 +515,13 @@ func (rm *resourceManager) newUpdateShardConfigurationRequestPayload(
 			res.SetReshardingConfiguration(shardsConfig)
 		}
 	} else if decrease {
+		// Compute shards to retain
+		if desiredSpec.NodeGroupConfiguration == nil {
+			for i := 0;  i < int(*desiredShardsCount); i++ {
+				shardsToRetain = append(shardsToRetain, desired.ko.Status.NodeGroups[i].NodeGroupID)
+			}
+		}
```
kumargauravsharma, line 523: Please consider moving this logic to an `else` block at line number 497. `shardsToRetain` is populated during if block and the else block would be logically connected to it.

kumargauravsharma (approved):

PR #62, opened 2021-09-30
Scale up/down and Scale in/out memory considerations

nmvk:
When spec is updated to perform Scale up and Scale in
we would need to perform Scale up first followed by
Scale in.

Similarly when we perform Scale down and Scale out
we would need to perform Scale out first.

### DIFF
--- a/pkg/resource/replication_group/custom_update_api.go
+++ b/pkg/resource/replication_group/custom_update_api.go
@@ -101,6 +101,19 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
 		return rm.decreaseReplicaCount(ctx, desired, latest)
 	}
 
+	// If there is a scale up modification, then we would prioritize it
+	// over increase/decrease shards. This is important since performing
+	// scale in without scale up might fail due to insufficient memory.
+	if delta.DifferentAt("Spec.CacheNodeType") && desired.ko.Status.AllowedScaleUpModifications != nil {
+		if desired.ko.Spec.CacheNodeType != nil {
+			for _, scaleUpInstance := range desired.ko.Status.AllowedScaleUpModifications {
+				if *scaleUpInstance == *desired.ko.Spec.CacheNodeType {
+					return nil, nil
+				}
+			}
+		}
+	}
+
 	// increase/decrease shards
 	if rm.shardConfigurationsDiffer(desired, latest) {
 		return rm.updateShardConfiguration(ctx, desired, latest)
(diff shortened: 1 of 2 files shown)


pkg/resource/replication_group/custom_update_api.go
```diff
@@ -101,6 +101,28 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
...
+	// scale in without scale up might fail due to insufficient memory.
+	if delta.DifferentAt("Spec.CacheNodeType") && desired.ko.Spec.ReplicationGroupID != nil {
+		input := &svcsdk.ListAllowedNodeTypeModificationsInput{}
+		input.SetReplicationGroupId(*desired.ko.Spec.ReplicationGroupID)
+		resp, apiErr := rm.sdkapi.ListAllowedNodeTypeModifications(input)
+		rm.metrics.RecordAPICall("READ_MANY", "ListAllowedNodeTypeModifications", apiErr)
+
+		if apiErr != nil {
+			return nil, apiErr
+		}
```
kumargauravsharma, line 115: The API call can be avoided here since https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/pkg/resource/replication_group/custom_set_output.go#L130-L138 populates the `ko.Status.AllowedScaleDownModifications` and `ko.Status.AllowedScaleUpModifications` into `latest` object.

pkg/resource/replication_group/custom_update_api_test.go
```diff
@@ -391,6 +392,67 @@ func TestCustomModifyReplicationGroup_NodeGroup_available(t *testing.T) {
...
+		var delta ackcompare.Delta
+		delta.Add("Spec.CacheNodeType", "cache.t3.small", "cache.m5.large")
+		delta.Add("Spec.NumNodeGroups", 3, 4)
+
+		var ctx context.Context
+		res, err := rm.CustomModifyReplicationGroup(ctx, desired, latest, &delta)
+		assert.Nil(res)
+		assert.Nil(err)
+		assert.NotEmpty(mocksdkapi.Calls)
+		assert.Equal("ListAllowedNodeTypeModifications", mocksdkapi.Calls[0].Method)
```
kumargauravsharma, line 422: Please add an assert statement to confirm there is no call to `decrease-replica-count` API.

nmvk: Update to assert empty `SDK` calls

pkg/resource/replication_group/custom_update_api.go
nmvk: Thank you! this simplifies the implementation and avoids additional API call.

kumargauravsharma (approved):

PR #67, opened 2021-10-12
Declarative Tests for ScalingUpAndDown and Async rollback fixes

nmvk:
Declarative tests for ScaleUpAndDown and fixes for Async rollback.

Upon Async rollback we would set the terminal condition to notify the user about the rollback. This would allow user to take the necessary action.

```
 Conditions:
    Last Transition Time:  2021-11-23T11:45:48Z
    Message:               replication group currently being modified. cannot delete.
    Status:                False
    Type:                  ACK.ResourceSynced
    Message:               InvalidParameterCombination: Cannot update CacheNodeType, Refer to Events for more details
    Status:                True
    Type:                  ACK.Terminal
```


test/e2e/scenarios/AddRemoveReplica/cmd_decrease_replica.yaml
```diff
@@ -0,0 +1,44 @@
+id: "CMD_DECREASE_REPLICA"
+description: "Decrease replica count for CMD"
+resource:
+  apiVersion: $CRD_GROUP/$CRD_VERSION
+  kind: ReplicationGroup
+  metadata:
+    name: test$RANDOM_SUFFIX
```
kumargauravsharma, line 7: If the name matches the scenario, it helps in debugging using cli later in case of failure.
Please consider names such as: `cmd-dec-rep-$RANDOM_SUFFIX` for this scenario

test/e2e/scenarios/AddRemoveReplica/cmd_decrease_replica_multi_az.yaml
```diff
@@ -0,0 +1,54 @@
...
+        replicationGroupDescription: Decrease replica count for CMD with MultiAZ enable to disabled
+        cacheNodeType: cache.t3.micro
+        numNodeGroups: 1
+        multiAZ: enabled
+        nodeGroupConfiguration:
+          - nodeGroupID: "0001"
+            replicaCount: 1
+            primaryAvailabilityZone: "us-west-2b"
+            replicaAvailabilityZones:
+              - "us-west-2c"
```
kumargauravsharma, line 24: Consider supplying these as replacement parameters. It will help in running the scenarios for other Region, AZ with minimal changes.

pkg/resource/replication_group/custom_set_output.go
```diff
@@ -297,3 +304,54 @@ func (rm *resourceManager) setLastRequestedLogDeliveryConfigurations(
...
+// annotation of the object
+func (rm *resourceManager) setLastRequestedNodeGroupConfiguration(
+	r *resource,
+	ko *svcapitypes.ReplicationGroup,
+) {
+	desiredAnnotations := r.ko.ObjectMeta.GetAnnotations()
+	annotations := make(map[string]string)
+	for k, v := range desiredAnnotations {
+		annotations[k] = v
+	}
```
A-Hilaly, line 329: Is it necessary to copy the map every time we need to set one annotation? maybe we can just initiated ones and set the annotations?

pkg/resource/replication_group/custom_set_output.go
```diff
@@ -297,3 +304,54 @@ func (rm *resourceManager) setLastRequestedLogDeliveryConfigurations(
...
+// annotation of the object
+func (rm *resourceManager) setLastRequestedNumNodeGroups(
+	r *resource,
+	ko *svcapitypes.ReplicationGroup,
+) {
+	desiredAnnotations := r.ko.ObjectMeta.GetAnnotations()
+	annotations := make(map[string]string)
+	for k, v := range desiredAnnotations {
+		annotations[k] = v
+	}
```
A-Hilaly, line 350: ditto

pkg/resource/replication_group/custom_set_output.go
```diff
@@ -279,7 +286,7 @@ func (rm *resourceManager) setAnnotationsFields(
 	}
```
A-Hilaly, line 286: please run `gofmt` on this file

pkg/resource/replication_group/custom_update_api.go
```diff
@@ -70,6 +73,22 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
 			requeue.DefaultRequeueAfterDuration)
 	}
 
+	// Handle the asynchronous rollback case for while Scaling down.
+	// This means that we have already attempted to apply the CacheNodeType once and
+	// were not successful hence we will set a terminal condition.
+	if !cacheNodeTypeRequiresUpdate(desired) && delta.DifferentAt("Spec.CacheNodeType") {
+		return nil, awserr.New("InvalidParameterCombination", "Cannot update CacheNodeType, "+
```
A-Hilaly, line 80: Is there a reason why you are using `awserr.New` and not `errors.New` ?

pkg/resource/replication_group/custom_update_api.go
```diff
@@ -691,3 +721,40 @@ func (rm *resourceManager) newModifyReplicationGroupRequestPayload(
 
 	return input
 }
+
+// cacheNodeTypeRequiresUpdate retrieves the last requested cacheNodeType saved in annotations and compares them
```
A-Hilaly, line 725: if i understand well there will always be only on `cacheNodeType` saved in the `AnnotationLastRequestedCNT` and not multiple ones?

pkg/resource/replication_group/custom_update_api.go
```diff
@@ -691,3 +721,40 @@ func (rm *resourceManager) newModifyReplicationGroupRequestPayload(
...
+	return true
+}
+
+// nodeGroupRequiresUpdate retrieves the last applied NumNodeGroups and NodeGroupConfiguration and compares them
+// to the current desired NumNodeGroups and NodeGroupConfiguration
+func nodeGroupRequiresUpdate(desired *resource) bool {
+	annotations := desired.ko.ObjectMeta.GetAnnotations()
+
+	if val, ok := annotations[AnnotationLastRequestedNNG]; ok && val != "null" {
+		numNodes, _ := strconv.ParseInt(val, 10, 64)
```
A-Hilaly, line 743: nit: I personally think that we should not ignore errors in int/json parsing and return an error when it's the case.

pkg/resource/replication_group/custom_update_api_test.go
```diff
@@ -392,6 +393,56 @@ func TestCustomModifyReplicationGroup_NodeGroup_available(t *testing.T) {
...
+		desiredNodeGroup := int64(4)
+		currentNodeGroup := int64(3)
+		desired.ko.Annotations[AnnotationLastRequestedNNG] = strconv.Itoa(int(desiredNodeGroup))
+		desired.ko.Spec.NumNodeGroups = &desiredNodeGroup
+		rm := provideResourceManager()
+
+		var delta ackcompare.Delta
+		delta.Add("Spec.NumNodeGroups", currentNodeGroup, desiredNodeGroup)
+
+		var ctx context.Context
```
A-Hilaly, line 439: nit: i know that `CustomModifyReplicationGroup` doesn't touch the `ctx` variable here, but ideally we should use `context.TODO()`

pkg/resource/replication_group/custom_update_api.go
nmvk: This is correct

pkg/resource/replication_group/custom_update_api.go
nmvk: This handles setting terminal codes correctly. Terminal codes are set using `awsErr.Code()`

pkg/resource/replication_group/custom_set_output.go
```diff
@@ -76,6 +79,9 @@ func (rm *resourceManager) CustomModifyReplicationGroupSetOutput(
...
 	ko.Spec.LogDeliveryConfigurations = logDeliveryConfig
 
+	// Keep the value of desired for CacheNodeType.
+	ko.Spec.CacheNodeType = r.ko.Spec.CacheNodeType
```
kumargauravsharma, line 83: Should this be moved to `customSetOutput` instead?

pkg/resource/replication_group/custom_update_api.go
```diff
@@ -691,3 +727,44 @@ func (rm *resourceManager) newModifyReplicationGroupRequestPayload(
...
+			return true
+		}
+
+		return false
+	}
+
+	desiredNodeGroupConfig := desired.ko.Spec.NodeGroupConfiguration
+	if val, ok := annotations[AnnotationLastRequestedNGC]; ok && val != "null" {
+		_ = json.Unmarshal([]byte(val), &desiredNodeGroupConfig)
+		return !reflect.DeepEqual(desiredNodeGroupConfig, val)
```
kumargauravsharma, line 765: Could delta APIs be used here instead of reflect?

kumargauravsharma (approved):

pkg/resource/replication_group/custom_update_api.go
nmvk: We use the reflect in `delta.go`

kumargauravsharma:
e2e tests have passed. merging the changes.

PR #68, opened 2021-10-18
Fix cache parameter group test failure

echen-98:
Description of changes:

Associating then updating a cache parameter group to a replication group previously did not work. Now we populate `latest.Spec.CacheParameterGroupName` from the response of the `DescribeCacheClusters` API.


echen-98:
Fixed unit test failures and removed a unit test redundant with existing declarative unit tests. Reason for that being we can't necessarily expect ReadOne to result in no difference between `latest` and `desired` if default fields exist and those fields map back to `latest.Spec` (meaning `desired.Spec` would have `nil` and `latest.Spec` would have some non-nil default value).

pkg/resource/replication_group/post_set_output.go
```diff
@@ -88,3 +89,16 @@ func setMaintenanceWindow(
...
+// setCacheParameterGroup updates the cache parameter group associated with the replication group
+//   (if non-nil in API response) regardless of whether it was specified in desired
+func setCacheParameterGroup(
+	latestCacheCluster *svcsdk.CacheCluster,
+	resource *resource,
+) {
+	ko := resource.ko
+	if latestCacheCluster.CacheParameterGroup != nil && latestCacheCluster.CacheParameterGroup.CacheParameterGroupName != nil {
+		cpgName := *latestCacheCluster.CacheParameterGroup.CacheParameterGroupName
+		ko.Spec.CacheParameterGroupName = &cpgName
```
jaypipes, line 102: if `ko.Spec.CacheParameterGroupName` was non-nil and different from `latestCacheCluster.CacheParameterGroup.CacheParameterGroupName`, why would we overwrite `ko.Spec.CacheParameterGroupName` here? This smells funny...

echen-98: The intention is just to retrieve the cache parameter group currently associated to the replication group. That information comes from the `DescribeCacheClusters` API since it is not present in the `DescribeReplicationGroups` API response.

If `ko.Spec.CacheParameterGroupName` is non-nil and gets overwritten, that's precisely what we want because the original value was copied over from `desired`, the latest value is different from that, and the controller will go on to detect a diff and invoke a Modify call with the desired `CacheParameterGroupName`.

echen-98: https://github.com/aws-controllers-k8s/community/issues/846#issuecomment-868887930

This falls under "case 3" of spec fields as I mentioned in this issue a while back

PR #84, opened 2022-03-16
Remove 404s from terminal codes

RedbackThomson:
Fixes https://github.com/aws-controllers-k8s/community/issues/1228

Description of changes:
404 error codes should not be included as terminal errors, as they prevent proper resource referencing. The better behaviour is for the controller to fall off exponentially while attempting to retry until the resource becomes available. This produces more noise in the controller logs, but properly updates the resources with the conditions until the resource can finally be created.


nmvk (approved):
LGTM

A-Hilaly (approved):

PR #111, opened 2023-03-10
Add subnetIDs after custom describe to fix `spec.subnetIDs: Required value`

mustafasencer:
**Issue #, if available:** https://github.com/aws-controllers-k8s/community/issues/1725

**Description of changes:**
- Adding the subnetIDs after the custom describe operation in order to populate the `spec.subnetIDs` field.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #112, opened 2023-03-10
Fix `No modifications were requested` issue during sdkUpdate for CacheSubnetGroup

mustafasencer:
**Description of changes:**

- Encountered `InvalidParameterValue: No modifications were requested.` during `sdkUpdate` just after an adoption of an existing resource. I introduced a custom update for the `CacheSubnetGroup` resource in order to swallow the error during `sdkUpdate`. Waiting your comments/reviews in case there is a better approach to handle this issue.

**For more context:** https://github.com/aws-controllers-k8s/elasticache-controller/pull/111#pullrequestreview-1335150235


pkg/resource/cache_subnet_group/custom_update.go
```diff
@@ -0,0 +1,127 @@
...
+	rm.metrics.RecordAPICall("UPDATE", "ModifyCacheSubnetGroup", err)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Message() {
+			case "No modifications were requested.":
+				return desired, nil
+			}
+		}
+		return nil, err
+	}
```
A-Hilaly, line 42: IMO this is happening because the controller is trying to make an update request when it shouldn't. Idealy we should either fix the delta/readone first. And maybe eventually leverage [`OmitUnchangedField`](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/config/resource.go#L322-L326) to generate the update function

mustafasencer: Based on your suggestion, I investigated a bit more and found that the post set output template I added ([here](https://github.com/aws-controllers-k8s/elasticache-controller/pull/111)) was fixing the primary issue where the controller fails with a **Reconciler error** (`spec.subnetIDs: Required value`) before even starting the `Sync` and thus not hitting `sdkFind`. 
After the addition of a read many post set output template, now I see that the `Sync` starts with success and the template logic appends the `ko.Spec.SubnetIDs` slice with duplicate values which misleads the delta to which leads to updating the resource.

_My quick solution:_ I am just checking if `ko.Spec.SubnetIDs` is populated or not before proceeding.
_Another approach (perhaps):_ If there is another way to start `Sync` with success without a template addition, please advise, it could perhaps be a more sound approach 🤷

A-Hilaly: I think we should not have duplicates at all - we need to focus on properly appending the subnets. Check my comment in the other PR https://github.com/aws-controllers-k8s/elasticache-controller/pull/111/files#r1133856530
@mustafasencer great job on testing and investigating this. I see that there were no elasticate releases 2 days ago but you still test the controllers - I'm curious to know how you do it :)

mustafasencer: @A-Hilaly Thank for the prompt response! I addressed the comment in the new commit.

I am testing locally via the kind cluster where I build only the local image of the controller and test the adopted resource scenario.

PR #115, opened 2023-03-28
redis >=6 skip patch version compare

joaquinservetto:
Issue #, if available:

Description of changes:

The purpose of this changes is to avoid the controller to compare patch versions when Redis version is >=6, since It's managed by AWS. More details [here](https://github.com/aws-controllers-k8s/community/issues/1737). 
 
### DIFF
--- a/pkg/resource/replication_group/delta_util.go
+++ b/pkg/resource/replication_group/delta_util.go
@@ -83,21 +83,14 @@ func engineVersionsMatch(
 		return true
 	}
 
-	// if the last character of desiredEV is "x", only check for a major version match
+	dMaj, dMin, _ := versionNumbersFromString(desiredEV)
+	lMaj, lMin, _ := versionNumbersFromString(latestEV)
 	last := len(desiredEV) - 1
-	if desiredEV[last:] == "x" {
-		// cut off the "x" and replace all occurrences of '.' with '\.' (as '.' is a special regex character)
-		desired := strings.Replace(desiredEV[:last], ".", "\\.", -1)
-		r, _ := regexp.Compile(desired + ".*")
-		return r.MatchString(latestEV)
-	}
 
-	// if the version is higher than 6, skip the upstream patch version when comparing.
+	// if the last character of desiredEV is "x" or the major version is higher than 5, ignore patch version when comparing.
 	// See https://github.com/aws-controllers-k8s/community/issues/1737
-	majorVersion, _ := strconv.Atoi(desiredEV[0:1])
-	if majorVersion >= 6 {
-		r, _ := regexp.Compile(desiredEV + ".*")
-		return r.MatchString(latestEV)
+	if dMaj > 5 || desiredEV[last:] == "x" {
+		return dMaj == lMaj && (dMin < 0 || dMin == lMin)
 	}
 
 	return false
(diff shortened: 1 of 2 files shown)


pkg/resource/replication_group/delta_util.go
```diff
@@ -91,6 +92,13 @@ func engineVersionsMatch(
 		return r.MatchString(latestEV)
 	}
```
Julian-Chu, line 94: would be great to add link issue in the comment,  i.e.:  
```golang
// See https://github.com/aws-controllers-k8s/community/issues/1737
````

**Review by Julian-Chu — COMMENTED** [review]
looks good, some small suggestions.

**Comment by Julian-Chu** [comment]
link to issue https://github.com/aws-controllers-k8s/community/issues/1737

File: pkg/resource/replication_group/delta_util.go [hunk 2]
```diff
@@ -91,6 +92,14 @@ func engineVersionsMatch(
...
 
+	// if the version is higher than 6, skip the upstream patch version when comparing.
+	// See https://github.com/aws-controllers-k8s/community/issues/1737
+	majorVersion, _ := strconv.Atoi(desiredEV[0:1])
+	if majorVersion >= 6 {
+		r, _ := regexp.Compile(desiredEV + ".*")
```
A-Hilaly, line 99: regex compile is an expensive operation i wouldn't recommend calling everytime the reconciliation loop is called. How about declaring it outside?
Maybe even better, how about using a semver parsing library and just make sure major/minor versions match?

A-Hilaly: I just saw that `regex.Compare` is above as well... ideally we should not.. (git blame 👀 )

pkg/resource/replication_group/delta_util.go
```diff
@@ -91,6 +92,14 @@ func engineVersionsMatch(
...
+		return r.MatchString(latestEV)
+	}
+
```
jaypipes, line 102: Instead of using regexes, how about just checking the major version of desired and latest? You can also add a tiny helper function to get the major version number from the string in a stable, safe way:

```go
// versionNumbersFromString takes a version string like "6.2", "6.x" or "7.0.4" and
// returns the major, minor and patch numbers. If no minor or patch numbers are present
// or contain the "x" placeholder, -1 is returned for that version number.
func versionNumbersFromString(version string) (int, int, int) {
    parts := strings.Split(version, ".")
    major := -1
    minor := -1
    patch := -1
    if len(parts) == 0 {
        return major, minor, patch
    }
    major, _ = strconv.Atoi(parts[0])
    if len(parts) > 1 {
        if !strings.EqualFold(parts[1], "x") {
            minor, _ = strconf.Atoi(parts[1])
        }
    }
    if len(parts) > 2 {
        if !strings.EqualFold(parts[2], "x") {
            patch, _ = strconf.Atoi(parts[2])
        }
    }
    return major, minor, patch
}
```

and in the `engineVersionsMatch()` function you can do something like this:

```go
dMaj, dMin, dPatch := versionNumbersFromString(desiredEV)
lMaj, lMin, lPatch := versionNumbersFromString(latestEV)
if dMaj > 5 {
    // if the version is higher than 5, ignore patch version when comparing.
    return dMaj == lMaj && (dMin < 0 || dMin == lMin)
}
```

jaypipes (changes requested):
thank you for your contribution @joaquinservetto! Please do add a PR description that includes the GH issue identifier (aws-controllers-k8s/community#1737) and a summary of why you are making these changes.

Also left a couple notes inline. Thank you!

joaquinservetto:
Thanks for your reviews! @jaypipes, I included the function that you suggested and I also joined the conditional with the one for the "x" character, since this logic should work for both scenarios. Please take a look again

A-Hilaly (approved):
+++

jaypipes (approved):
thanks @joaquinservetto! :)

PR #117, opened 2023-05-01
Update template to support watch namespace

rahtr:
As the controller runtime supports watching namespaces, we would need to add a support to watch namespaces other than where the controller is deployed. This is a common use case for multi-tenant K8s environments. This PR introduces a `watchNamespace` flag which lets user define the namespaces to be watched by the controller. If left blank, it would take the release namespace.


jaypipes (commented):
As mentioned in the rds-controller PR, this will need to be done against the code generator repository...

PR #136, opened 2024-06-13
Fix horizontal scaling for ReplicationGroups

gfrey:
This will set the value for the number of node groups of a ReplicationGroup to the value given in the Spec, thereby enabling horizontal scaling.

Fixes aws-controllers-k8s/community#2080.

### DIFF
--- a/pkg/resource/replication_group/post_set_output.go
+++ b/pkg/resource/replication_group/post_set_output.go
@@ -89,6 +89,10 @@ func setNodeGroupConfiguration(
 
 		ko.Spec.NodeGroupConfiguration = nodeGroupConfigurations
 	}
+
+	if respRG.NodeGroups != nil && ko.Spec.NumNodeGroups != nil {
+		*ko.Spec.NumNodeGroups = int64(len(respRG.NodeGroups))
+	}
 }
 
 //TODO: for all the fields here, reevaluate if the latest observed state should always be populated,


gfrey:
I forced push an updated commit message (I had vertical and horizontal scaling mixed up).

gfrey:
Converted to a draft, as I can't properly verify the fix. Even without it the [test_rg_scale_horizontally test](https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/test/e2e/tests/test_replicationgroup.py#L730) works just fine. When I noticed the issue, I had updated the `NumNodeGroup` in the spec, but that didn't trigger any changes to the replication group. Only when I also updated the instance type, the horizontal scaling was also picked up.

Let me first try to recreate the issue.

gfrey:
I've tested this with a resource like the following:
```
---
apiVersion: "elasticache.services.k8s.aws/v1alpha1"
kind: ReplicationGroup
metadata:
  name: my-test
  namespace: default
spec:
  description: CloudCache Redis my-test
  automaticFailoverEnabled: true
  engine: Redis
  engineVersion: "7.0"
  cacheNodeType: cache.t4g.micro
  replicasPerNodeGroup: 1
  port: 6379
  multiAZEnabled: true
  replicationGroupID: my-test-cloudcache
  cacheParameterGroupName: default.redis7.cluster.on

  numNodeGroups: 1

  atRestEncryptionEnabled: true
  transitEncryptionEnabled: true
  preferredMaintenanceWindow: "sun:22:00-sun:23:00"
```

1. Using the controller from the main branch, a change to `numNodeGroups` doesn't start updating the resource. Subsequent updates to other fields (say the instance type) will also trigger the horizontal scaling.
2. Using the controller from this PR, the change will be detected and trigger the expected update.

I'm not sure, why the `k8s.patch_custom_resource` call from `test_rg_scale_horizontally` triggers an update.

pkg/resource/replication_group/post_set_output.go
```diff
@@ -89,6 +89,10 @@ func setNodeGroupConfiguration(
 
 		ko.Spec.NodeGroupConfiguration = nodeGroupConfigurations
 	}
+
+	if respRG.NodeGroups != nil && ko.Spec.NumNodeGroups != nil {
+		*ko.Spec.NumNodeGroups = int64(len(respRG.NodeGroups))
+	}
```
a-hilaly, line 95: Nice catch!

PR #137, opened 2024-06-28
Support for CacheCluster resource

cPu1:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/2079

Description of changes:

This changelist adds support for the ElastiCache `CacheCluster` resource. The generated code did not account for many fields in the call to `ModifyCacheCluster`, and the `DescribeCacheClusters` API does not return all fields in the spec and values for some fields are reported in the `PendingModifiedValues` field instead. To handle this, the last requested configuration for `Spec.PreferredAvailabilityZones` is stored as annotations and used to compute `NewAvailabilityZones`. Moreover, `CacheNodeIdsToRemove` is computed using `Spec.NumCacheNodes` and `PendingModifiedValues.NumCacheNodes`. Support for `Spec.LogDeliveryConfigurations` is explicitly removed and parked for later because the Describe API does not return all fields for this struct. There were many changes to other ElastiCache resources introduced in the newer SDK, so they have been removed for now and not ignored. As a result, a subsequent `make build-controller` will result in new changes.

The logic for syncing tags and comparing engine versions has been refactored to promote reusability.

Closes https://github.com/aws-controllers-k8s/community/issues/2079.

Signed-off-by: cPu1 <patwal.chetan@gmail.com>


templates/hooks/cache_cluster/sdk_update_post_set_output.go.tpl
```diff
@@ -0,0 +1,16 @@
...
+		}
+		if pendingModifications.TransitEncryptionEnabled != nil {
+			ko.Spec.TransitEncryptionEnabled = pendingModifications.TransitEncryptionEnabled
+		}
+	}
+	if err == nil {
+		// Setting resource synced condition to false will trigger a requeue of
+		// the resource. No need to return a requeue error here.
+		ackcondition.SetSynced(&resource{ko}, corev1.ConditionFalse, &condMsgCurrentlyUpdating, nil)
+	}
```
a-hilaly, line 16: I think that we do not need to requeue after an Update api call.

test/e2e/tests/test_cache_cluster.py
```diff
@@ -0,0 +1,210 @@
...
+
+def get_and_assert_status(ref: k8s.CustomResourceReference, expected_status: str, expected_synced: bool):
+    cr = k8s.get_resource(ref)
+    assert cr is not None
+    assert 'status' in cr
+
+    logging.info("---status")
+    logging.info(cr['status']['cacheClusterStatus'])
+    print(cr)
+    print(cr['status'])
```
a-hilaly, line 67: forgotten logs

test/e2e/tests/test_cache_cluster.py
a-hilaly, line 1: nice testing

pkg/util/engine_version.go
```diff
@@ -0,0 +1,69 @@
...
+// or contain the "x" placeholder, -1 is returned for that version number.
+func versionNumbersFromString(version string) (int, int, int) {
+	parts := strings.Split(version, ".")
+	major := -1
+	minor := -1
+	patch := -1
+	if len(parts) == 0 {
+		return major, minor, patch
+	}
+	major, _ = strconv.Atoi(parts[0])
```
a-hilaly, line 57: need to handle the error here - if a user provides a completely wrong version we need to set the resource ona terminal condition stating that the engine version format is incorrect.

pkg/util/engine_version_test.go
a-hilaly, line 1: 👍

pkg/util/engine_version.go
```diff
@@ -0,0 +1,69 @@
...
+		if !strings.EqualFold(parts[1], "x") {
+			minor, _ = strconv.Atoi(parts[1])
+		}
+	}
+	if len(parts) > 2 {
+		if !strings.EqualFold(parts[2], "x") {
+			patch, _ = strconv.Atoi(parts[2])
+		}
+	}
+	return major, minor, patch
```
a-hilaly, line 68: looks like the `patch` variable is never used

pkg/util/engine_version.go
```diff
@@ -0,0 +1,69 @@
...
+		return true
+	}
+
+	dMaj, dMin, _ := versionNumbersFromString(desiredEV)
+	lMaj, lMin, _ := versionNumbersFromString(latestEV)
+	last := len(desiredEV) - 1
+
+	// if the last character of desiredEV is "x" or the major version is higher than 5, ignore patch version when comparing.
+	// See https://github.com/aws-controllers-k8s/community/issues/1737
+	if dMaj > 5 || desiredEV[last:] == "x" {
```
a-hilaly, line 39: qq: why 5? what if a user provides `4.x`

pkg/resource/cache_cluster/hooks.go
```diff
@@ -0,0 +1,133 @@
...
+				input.CacheNodeIdsToRemove = append(input.CacheNodeIdsToRemove, &nodeID)
+			}
+		}
+	}
+
+	if idx := slices.IndexFunc(delta.Differences, func(diff *ackcompare.Difference) bool {
+		return diff.Path.Contains("Spec.PreferredAvailabilityZones")
+	}); idx != -1 && desired.ko.Spec.PreferredAvailabilityZones != nil {
+		if nodesDelta >= 0 {
+			return errors.New("spec.preferredAvailabilityZones can only be changed when new nodes are being added via spec.numCacheNodes")
```
a-hilaly, line 119: do you want to make those terminal errors?

pkg/resource/cache_cluster/hooks.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		oldAZsLen := 0
+		oldValues, ok := delta.Differences[idx].B.([]*string)
+		if ok {
+			oldAZsLen = len(oldValues)
+		}
+		if len(desiredSpec.PreferredAvailabilityZones) <= oldAZsLen {
+			return errors.New("newly specified AZs in spec.preferredAvailabilityZones must match the number of cache nodes being added")
```
a-hilaly, line 128: same here

pkg/resource/cache_cluster/hooks.go
```diff
@@ -0,0 +1,133 @@
...
+		numNodes := *latest.ko.Spec.NumCacheNodes
+		if pendingModifications := latest.ko.Status.PendingModifiedValues; pendingModifications != nil &&
+			pendingModifications.NumCacheNodes != nil && *pendingModifications.NumCacheNodes > numNodes {
+			numNodes = *pendingModifications.NumCacheNodes
+		}
+		nodesDelta = numNodes - *desired.ko.Spec.NumCacheNodes
+		if nodesDelta > 0 {
+			for i := numNodes; i > numNodes-nodesDelta; i-- {
+				nodeID := fmt.Sprintf("%04d", i)
+				input.CacheNodeIdsToRemove = append(input.CacheNodeIdsToRemove, &nodeID)
```
a-hilaly, line 110: This is a very weird API...

a-hilaly (commented):
@cPu1 thank you so much! this is great. I left few comments in-line

pkg/util/engine_version.go
a-hilaly: e.g https://github.com/aws-controllers-k8s/eks-controller/blob/78062e26cdc53435c0dae3913d959be6514d0f5f/pkg/resource/cluster/hook.go#L329

cPu1: This wasn't introduced in this PR, the function was merely moved from `pkg/resource/replication_group/delta_util.go`. If it's changed to return an error, the ReplicationGroup controller will also need to be refactored.

pkg/util/engine_version.go
cPu1: This function was also moved from `pkg/resource/replication_group/delta_util.go`.

pkg/util/engine_version.go
cPu1: This was also moved from `pkg/resource/replication_group/delta_util.go`. The reason is mentioned in this PR: https://github.com/aws-controllers-k8s/elasticache-controller/pull/115. 

> The purpose of this changes is to avoid the controller to compare patch versions when Redis version is >=6, since It's managed by AWS. More details https://github.com/aws-controllers-k8s/community/issues/1737.

pkg/resource/cache_cluster/hooks.go
cPu1: It's a terminal error. Any error returned from this function is treated as a terminal error.

templates/hooks/cache_cluster/sdk_update_post_set_output.go.tpl
cPu1: I checked the generated code. I think we don't need to have this here.

a-hilaly (approved):
Thank you @cPu1 ! 👍 
/lgtm

PR #175, opened 2025-07-17
Add serverlesscache to elasticache

rushmash91:
fixes https://github.com/aws-controllers-k8s/community/issues/2522

Description of changes:
- Add support serverlesscache resource


rushmash91:
unrelated test..
/test elasticache-kind-e2e

apis/v1alpha1/generator.yaml
```diff
@@ -296,7 +296,62 @@ resources:
         - TagQuotaPerResourceExceeded
     update_operation:
       custom_method_name: customUpdateUserGroup
+  ServerlessCache:
+    update_operation:
+      custom_method_name: customUpdateServerlessCache
```
michaelhtm, line 301: I don't think a hook is necessary for update...
https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_ModifyServerlessCache.html
Code-gen should be able to generate this

apis/v1alpha1/generator.yaml
```diff
@@ -296,7 +296,62 @@ resources:
...
+    fields:
+      ServerlessCacheName:
+        is_primary_key: true
```
michaelhtm, line 304: rename to `Name`
not sure if you need to define the `is_primary_key` here

apis/v1alpha1/generator.yaml
```diff
@@ -296,7 +296,62 @@ resources:
...
+          path: Status.SubnetID
+      UserGroupIDs:
+        references:
+          resource: UserGroup
+          path: Spec.UserGroupID
+    synced:
+      when:
+      - path: Status.Status
+        in:
+        - available
```
michaelhtm, line 323: The status here should be capitalized: https://github.com/aws/aws-sdk-go-v2/blob/0475b822f1f48ca746369818d51272f989c06b91/service/elasticache/types/types.go#L2019

Also, should CREATE_FAILED be a synced/terminal condition?

apis/v1alpha1/generator.yaml
```diff
@@ -296,7 +296,62 @@ resources:
...
+    exceptions:
+      terminal_codes:
+        - ServerlessCacheAlreadyExistsFault
+        - ServerlessCacheQuotaForCustomerExceededFault
+        - InvalidParameterValue
+        - InvalidParameterCombination
+        - InvalidVPCNetworkStateFault
+        - TagQuotaPerResourceExceeded
+        - InvalidKMSKeyFault
```
michaelhtm, line 332: maybe we should ensure these are truly terminal

apis/v1alpha1/generator.yaml
```diff
@@ -296,7 +296,62 @@ resources:
...
+    hooks:
+      sdk_read_many_post_set_output:
+        template_path: hooks/serverless_cache/sdk_read_many_post_set_output.go.tpl
+    print:
+      add_age_column: true
+      add_synced_column: true
```
michaelhtm, line 338: these are true by default i believe

apis/v1alpha1/generator.yaml
```diff
@@ -296,7 +296,62 @@ resources:
...
+      order_by: index
+      additional_columns:
+      - name: STATUS
+        json_path: .status.status
+        type: string
+        index: 10
+      - name: ENDPOINT
+        json_path: .status.endpoint.address
+        type: string
+        index: 20
```
michaelhtm, line 348: Does this actually work? Would it be better to do something similar to this? https://github.com/aws-controllers-k8s/ec2-controller/blob/ef9cc69d5c0c750e1d75865268c99c88ac4b0365/generator.yaml#L968-L971

apis/v1alpha1/generator.yaml
```diff
@@ -296,7 +296,62 @@ resources:
...
 operations:
+  DescribeServerlessCaches:
+    operation_type:
+    - List
+    resource_name:
+      ServerlessCache
```
michaelhtm, line 354: was this not being caught by code-gen?

pkg/resource/serverless_cache/hooks.go
```diff
@@ -0,0 +1,305 @@
...
+	_, err := rm.sdkapi.ModifyServerlessCache(ctx, input)
+	rm.metrics.RecordAPICall("UPDATE", "ModifyServerlessCache", err)
+	return err
+}
+
+func isServerlessCacheCreating(r *resource) bool {
+	if r.ko.Status.Status == nil {
+		return false
+	}
+	return *r.ko.Status.Status == "creating"
```
michaelhtm, line 87: Maybe make the states Constants?
Also this one should also be capitalized

pkg/resource/serverless_cache/hooks.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func isServerlessCacheDeleting(r *resource) bool {
+	if r.ko.Status.Status == nil {
+		return false
+	}
+	return *r.ko.Status.Status == "deleting"
```
michaelhtm, line 94: ditto

pkg/resource/serverless_cache/hooks.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func isServerlessCacheModifying(r *resource) bool {
+	if r.ko.Status.Status == nil {
+		return false
+	}
+	return *r.ko.Status.Status == "modifying"
```
michaelhtm, line 101: ditto

pkg/resource/serverless_cache/hooks.go
```diff
@@ -0,0 +1,305 @@
...
+			return nil, fmt.Errorf("cannot update serverless cache: %v", err)
+		}
+
+	case delta.DifferentAt("Spec.CacheUsageLimits"):
+		if err := rm.syncCacheUsageLimits(ctx, desired); err != nil {
+			return nil, fmt.Errorf("cannot update serverless cache: %v", err)
+		}
+
+	case delta.DifferentAt("Spec.Engine") || delta.DifferentAt("Spec.MajorEngineVersion"):
+		if err := rm.syncEngineAndVersion(ctx, desired); err != nil {
```
michaelhtm, line 173: can we not set all values in the same call (except Tags)? with this code we will be updating a single field at a time

apis/v1alpha1/generator.yaml
rushmash91: Thank is inconsistent with this controller, ref other resources..

apis/v1alpha1/generator.yaml
rushmash91: the API does not allow changes to more than one field, this is a behavior similar to DynamoDB

apis/v1alpha1/generator.yaml
rushmash91: Should we change for all resources? havent had any customer issues reported yet.

pkg/resource/serverless_cache/hooks.go
rushmash91: That is the point 🙂

rushmash91: https://github.com/hashicorp/terraform-provider-aws/issues/35897

apis/v1alpha1/generator.yaml
rushmash91: https://github.com/hashicorp/terraform-provider-aws/issues/35897

apis/v1alpha1/generator.yaml
rushmash91: Yes they work. Just maintaining consistency with `cachecluster` as much as possible

PR #185, opened 2025-08-22
add node details to status

rushmash91:
Description of changes:

The AWS ElastiCache DescribeCacheClusters API requires the `ShowCacheNodeInfo` parameter to be explicitly set to `true` to return cache node details including endpoints. The ACK controller was calling this API without this parameter, resulting in responses that contained cluster metadata but omitted the node details like endpoints.

Added a direct modification to the sdkFind function in `pkg/resource/cache_cluster/sdk.go` to set `ShowCacheNodeInfo = true` on all `DescribeCacheClusters` API calls. 


a-hilaly (approved):
nice catch! @rushmash91 

/lgtm

PR #211, opened 2026-04-02
feat: support for IAM authentication mode for user

stephensarabiasurion:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/2356

Description of changes:

- Removed AuthenticationMode field_paths from the ignore list
- Updated the AuthenticationMode struct to pass in Type, in line with the ElastiCache CDK docs
- New authenticationMode block in helm CRDs
- Also added tests and doc for IAM authentication mode


jlbutler:
Hi @stephensarabiasurion thanks so much for the PR! 

Looks like some of the hooks aren't wired up properly. code-gen should handle that for you.

Did you hand-edit files for this PR? ACK leverages code generation, so you should not need to edit files like sdk.go, types.go, etc

We have contributor docs on the old docset, we prob need to update this.
https://aws-controllers-k8s.github.io/community/docs/contributor-docs/overview/

Can you try a reset and:

1. Update your ignores in generator.yaml (don't worry about apis/v1alpha1/generator.yaml, it gets built)
2. add your AuthenticationMode handling to pkg/resource/user/hooks.go
3. Rebuild — SERVICE=elasticache make -C code-generator build-controller

That should create the code, wire things up, and create your kustomize/helm artifacts for you. If all goes well, force push and we'll see where that lands? Holler if any of this doesn't make sense, thanks!

jlbutler:
@stephensarabiasurion I'm working on updating our contributor docs, in the meantime let us know if you'd like us to pick this up and finish up. Happy either way, just holler back if you'd like to take another swing or pass it off. Thanks!

stephensarabiasurion:
> @stephensarabiasurion I'm working on updating our contributor docs, in the meantime let us know if you'd like us to pick this up and finish up. Happy either way, just holler back if you'd like to take another swing or pass it off. Thanks!

@jlbutler sorry for the delay, was out sick. I'm picking this back up today, your docs make sense! I'll reach out if I run into anything crazy

stephensarabiasurion:
@jlbutler rebased, built with code-generator v0.58.0, and ran local tests again. i think it's ready for retest, let me know if i missed anything else! and thanks

apis/v1alpha1/types.go
```diff
@@ -37,6 +37,7 @@ type Authentication struct {
 // Specifies the authentication mode to use.
 type AuthenticationMode struct {
 	Passwords []*string `json:"passwords,omitempty"`
```
michaelhtm: Can we mark this field as `is_secret` in generator.yaml?
Eg. https://github.com/aws-controllers-k8s/secretsmanager-controller/blob/main/generator.yaml#L26

config/crd/bases/elasticache.services.k8s.aws_users.yaml
```diff
@@ -44,6 +44,16 @@ spec:
...
                   Regex Pattern: `\S`
                 type: string
+              authenticationMode:
+                description: Specifies how to authenticate the user.
+                properties:
+                  passwords:
+                    items:
+                      type: string
+                    type: array
+                  type_:
```
michaelhtm: can we change the goTag for `type_` to be `type`? eg. https://github.com/aws-controllers-k8s/glue-controller/blob/main/generator.yaml#L49-L50

pkg/resource/user/hooks.go
```diff
@@ -95,6 +96,14 @@ func (rm *resourceManager) populateUpdatePayload(
 		input.AccessString = r.ko.Spec.AccessString
 	}
 
+	if delta.DifferentAt("Spec.AuthenticationMode") && r.ko.Spec.AuthenticationMode != nil {
+		authMode := &svcsdktypes.AuthenticationMode{}
+		if r.ko.Spec.AuthenticationMode.Type != nil {
+			authMode.Type = svcsdktypes.InputAuthenticationType(*r.ko.Spec.AuthenticationMode.Type)
+		}
+		input.AuthenticationMode = authMode
```
michaelhtm: Should we handle Passwords as well here?

pkg/resource/user/sdk.go
michaelhtm, line 1: Looks like sdkFind is not retrieving the latest AuthenticationMode. In this case, updates will not work.
DescribeUsers returns `Authentication`, while in `CreateUser`, the input accepts AuthenticationMode. 

We may need to use `generator.yaml` `renames` field to trick the code-generator here.
https://aws-controllers-k8s.github.io/community/docs/contributor-docs/code-generator-config/#renames-renaming-things

https://aws-controllers-k8s.github.io/community/docs/contributor-docs/code-generator-config/#renaming-things-with-different-names-in-input-and-output-shapes

michaelhtm (commented):
Great work @stephensarabiasurion!
left a few comments below

config/crd/bases/elasticache.services.k8s.aws_users.yaml
stephensarabiasurion: have that change pending; but FYI, there's a status.authentication.type_ field also in here. i'm tempted to go ahead and update a go_tag override there too for consistency, but this could be a breaking change for some existing users. thoughts?

apis/v1alpha1/types.go
stephensarabiasurion: updated the generator.yaml User.fields section for authmode to `is_secret`

config/crd/bases/elasticache.services.k8s.aws_users.yaml
stephensarabiasurion: update pushed for authmode.type (left status alone)

pkg/resource/user/hooks.go
stephensarabiasurion: moved populateUpdatePayload to template_path for error handling, as do other elements in hooks.go

pkg/resource/user/sdk.go
stephensarabiasurion: created hooks for sdk_read_many_post_set_output, sdk_create_post_set_output, and sdk_update_post_set_output

stephensarabiasurion:
@michaelhtm submitted changes that i believe address your comments. one thing, on the renames:
i think a direct rename would replace spec.authenticationMode on every read (clearing the password), and would remove the status.authentication field which is a breaking change (also noted in-line on one of the other comments). template hooks were used on the other object types, so i'm hoping you agree. happy to try a different method if not/if i've overlooked something

stephensarabiasurion:
hi @jlbutler @michaelhtm , i'm running through local tests per ACK testing docs, and i think this is good. but wondering if there's a way i can run some kind of official test to see if this satisfies your requests for change, etc?

jlbutler:
@stephensarabiasurion that verify-code-gen failure is likely due to checking out a specific version of code-gen and runtime in your build environment. the CI job will checkout HEAD for both of those. can you rebuild with code-gen and runtime both at HEAD (pull first if you haven't in a while) and force push? that should sort that one. 

the tests ran clean, though i have a few comments on making them a bit more robust. will loop back here in a bit, thanks! EDIT: michael beat me to it, thanks again for taking the time to work through this!

config/crd/bases/elasticache.services.k8s.aws_users.yaml
michaelhtm: Thanks @stephensarabiasurion 
You're right, it may be too late to fix the status field without breaking it for existing users

templates/hooks/user/sdk_read_many_post_set_output.go.tpl
```diff
@@ -0,0 +1,8 @@
+	rm.setSyncedCondition(resp.Users[0].Status, &resource{ko})
```
michaelhtm: can we place this under some guard?
Also, why do we want to set synced condition here?

pkg/resource/user/hooks.go
michaelhtm: Thanks.
Btw, i just noticed we already set AuthenticationMode unconditionally in sdk.go newUpdateRequestPayload
Is there a reason we want this to be behind a `DifferentAt` check?

pkg/resource/user/hooks.go
```diff
@@ -87,20 +88,43 @@ func (rm *resourceManager) CustomModifyUser(
...
 	input *svcsdk.ModifyUserInput,
 	r *resource,
 	delta *ackcompare.Delta,
-) {
+) error {
 	if delta.DifferentAt("Spec.AccessString") && r.ko.Spec.AccessString != nil {
 		input.AccessString = r.ko.Spec.AccessString
 	}
 
+	if delta.DifferentAt("Spec.AuthenticationMode") && r.ko.Spec.AuthenticationMode != nil {
```
michaelhtm: Is there a way of unsetting/removing AuthenticationMode if r.ko.Spec.AuthenticationMode is nil?

michaelhtm (commented):
Thanks again @stephensarabiasurion 
left a few more comments

test/e2e/tests/test_user.py
```diff
@@ -128,6 +154,18 @@ def test_user_nopass(self, user_nopass, user_nopass_input):
...
+        (reference, resource) = user_iam
+        assert k8s.get_resource_exists(reference)
+
+        assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "True", wait_periods=5)
+        resource = k8s.get_resource(reference)
+        assert resource["status"]["lastRequestedAccessString"] == user_iam_input["ACCESS_STRING"]
+        assert resource["status"]["authentication"] is not None
+        assert resource["status"]["authentication"]["type_"] == "iam"
+        assert resource["spec"]["authenticationMode"]["type"] == "iam"
+
```
michaelhtm, line 174: can we also verify these values with the ones set in AWS? using a describe call?
I see it's not something that exists in this file, but it's probably something we can improve :)

templates/hooks/user/sdk_read_many_post_set_output.go.tpl
stephensarabiasurion: wrapped it in a len check for readability, another commit coming in.

fwiw, i think we're missing guards on a few other files (e.g. `templates/hooks/replication_group/sdk_read_many_post_set_output.go.tpl`), but i assume it was skipped because the sdkFind loop returns `ackerr.NotFound` before this. still nice to read it separately, though, for sure. i do actually have another change i'd like to bring to you all on replicationGroup, so maybe we can handle that one then

setSyncedCondition was originally in the inline hook code:
```
sdk_read_many_post_set_output:
    code: "rm.setSyncedCondition(resp.Users[0].Status, &resource{ko})"
```
so i just preserved it in this template file version

pkg/resource/user/hooks.go
stephensarabiasurion: i remember having set this to match the other fields' patterns, so that it's more vertically aligned i guess, but it is quite redundant sorry. removing

pkg/resource/user/hooks.go
stephensarabiasurion: i'm thinking we should not, but i'm interested if you feel differently:
the elasticache API doesn't unset the type if the API call lacks type, and our `user/sdk_read_many_post_set_output` always adds the existing type. it also seems that the API requires a specific authentication mechanism, and so derives an authenticationMode despite it showing as not required in docs

test/e2e/tests/test_user.py
stephensarabiasurion: done; ran some tests in my cluster to confirm, status fields look as expected

stephensarabiasurion:
@michaelhtm thanks for your input; i added comments in-line, and re-ran tests with your suggestions (had some ECR access issues, heh, thus the long time between comment and commit). let me know what you think

test/e2e/tests/test_user.py
michaelhtm: Thanks for adding AWS validation to other tests as well :)

michaelhtm (approved):
Thanks for going above and beyond @stephensarabiasurion!
/lgtm

PR #215, opened 2026-05-26
fix: case-insensitive Engine comparison to prevent drift

### LINKED ISSUE #2695: [RDS controller]: constant loop of changing the db in the cloud
We are using ACK Controllers to manage our database resources, and we are observing a persistent issue: after a `DBinstance` is created, it regularly falls back into the configuring cycle.
In the controller logs we constantly see messages like:

```
{"level":"info","ts":"2025-11-17T08:05:49.709Z","logger":"ackrt","msg":"desired resource state has changed","kind":"DBInstance","namespace":"infra-production","name":"database-2025-11-09","account":"966321756598","role":"arn:aws:iam::966321756598:role/ack-ec2-controller-tooling","region":"eu-west-2","is_adopted":false,"generation":2,"diff":[{"Path":{"Parts":["Spec","AutoMinorVersionUpgrade"]},"A":null,"B":true},{"Path":{"Parts":["Spec","CACertificateIdentifier"]},"A":null,"B":"rds-ca-rsa2048-g1"},{"Path":{"Parts":["Spec","LicenseModel"]},"A":null,"B":"general-public-license"},{"Path":{"Parts":["Spec","MultiAZ"]},"A":null,"B":false},{"Path":{"Parts":["Spec","PreferredBackupWindow"]},"A":null,"B":"00:56-01:26"},{"Path":{"Parts":["Spec","PreferredMaintenanceWindow"]},"A":null,"B":"sat:23:48-sun:00:18"},{"Path":{"Parts":["Spec","StorageEncrypted"]},"A":null,"B":false},{"Path":{"Parts":["Spec","StorageThroughput"]},"A":null,"B":0},{"Path":{"Parts":["Spec","StorageType"]},"A":null,"B":"gp2"}]}
```

This behavior clearly doesn’t seem normal. While the controller repeatedly attempts to “configure” the instance, the database may become unavailable — which is unacceptable in production.
It would be very helpful to have guidance or a built-in mechanism to prevent this repeating reconfiguration loop.
Additionally, we see a related issue when `AutoMinorVersionUpgrade` is enabled:
the cloud database automatically upgrades to a newer minor version, while the version in the ACK manifest remains unchanged. After that, any further changes applied via the manifest get stuck, because the version mismatch prevents updates from proceeding.
Could you please clarify whether this is expected behavior, and whether there are recommended practices to avoid these situations?

michaelhtm:
Description of changes:
AWS normalizes Engine to lowercase ("redis") but users may specify
"Redis". Add case-insensitive comparison in modifyDelta to suppress
the false delta. Also add skip_incomplete_check to PreferredAvailabilityZone.

Resolves aws-controllers-k8s/community#2695
https://github.com/aws-controllers-k8s/community/issues/2695#issuecomment-3661139739

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2026-06-23T18:18:31Z"
+  build_date: "2026-06-26T00:26:59Z"
   build_hash: 2ae5d2cfadaa2a10b2ccb9e73a111b2a91c36642
-  go_version: go1.26.0
+  go_version: go1.26.4
   version: v0.60.0
-api_directory_checksum: 8c1c46c231242e776757404b654cc2ab451f86fc
+api_directory_checksum: 2f3d1db010a35182aa2820cea4c75551a0b56e2f
 api_version: v1alpha1
 aws_sdk_go_version: v1.41.0
 generator_config_info:
-  file_checksum: 546ff4359a1259012dc41c3b0a5069eda6534627
+  file_checksum: 86974b386411515be027b7646b61959ca09ac2cb
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -32,6 +32,10 @@ resources:
           path: Status.ID
       AuthToken:
         is_secret: true
+      Engine:
+        set:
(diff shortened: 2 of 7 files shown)


knottnt:
@michaelhtm is the PR description correct? The link issue is for the RDS controller and I don't see any changes related to a PreferredAvailabilityZone field.

michaelhtm:
@knottnt There are comments in the issue that mention delta issues in ACK in general

knottnt (approved):
/lgtm

PR #223, opened 2026-07-11
feat: add Durability field support for ReplicationGroup

n-bezawada:
Adds the `durability` field to ReplicationGroup. Durability gives Valkey 9.0+ clusters
WAL-based persistence and accepts `disabled`, `default`, `async` and `sync`.

Most of this diff is regenerated code. The hand-written part is `generator.yaml`, two hook
templates, and the tests.

### Why it needs custom handling

`ModifyReplicationGroup` echoes back the *old* durability. Ask for `sync` on an `async` group
and the response says `async`, with `Status: available` and an empty `PendingModifiedValues` —
so there's no signal the change is still settling.

Two things fall out of that:

- **Write responses can't populate the field.** `set: ignore: from` is scoped to Create and
  Update, so only `DescribeReplicationGroups` fills in `Spec.Durability`. The generated delta
  then compares what you asked for against what the service reports, which also means a
  durability changed behind the controller's back gets corrected.
- **Something has to force a re-read.** Otherwise the resource reports `Synced=True` while the
  value is still stale. A `sdk_update_post_set_output` hook sets synced false when durability
  was in the delta. A second hook drops `Durability` from the request when it wasn't, the same
  way this resource already handles `LogDeliveryConfigurations`.

`Durability` is late initialized, since there's no way to unset it through the API — removing it
from the spec writes the reported value back instead of diffing against `nil` forever.
`CacheParameterGroupName` needed the same treatment: the service assigns a default parameter
group when you don't ask for one, and that permanent difference kept the custom modify path
engaged, so the generated update never ran.

### What the service allows

Durable vs non-durable is fixed at creation, but you can move within either set:

| From | To | Result |
|---|---|---|
| `async` | `sync` (and back) | allowed |
| `default` | `disabled` (and back) | allowed |
| `default` / `disabled` | `async` / `sync` | `Durability cannot currently be enabled on non-durable clusters` |
| `async` / `sync` | `default` / `disabled` | `Durability cannot currently be disabled on durable clusters` |

Both codes are already in this resource's `terminal_codes`, so rejections surface as terminal
conditions on their own. No transition validation in the controller.

Creation also needs cluster mode, Multi-AZ and in-transit encryption — the service rejects the
request outright without them.

### Testing

The comparison is generated,  Five e2e cases in `test/e2e/tests/test_replicationgroup.py`: 

- `..._update` — create `async`, update to `sync`, checking spec, status and the AWS-side value
  at both ends. Also asserts the resource settles instead of flapping back into modifying.
- `..._drift` — change durability outside the controller, confirm the spec value is restored.
  The controller doesn't poll for drift, so the test edits an unrelated field to trigger a
  reconcile rather than waiting on the resync interval.
- `..._unset` — remove the field, confirm late initialization writes the value back and nothing
  goes terminal.
- `..._disabled_guard` / `..._default_guard` — the two rejected directions surface as terminal
  conditions carrying the service's own error, and leave the AWS-side resource alone.

### On the diff size

Bigger than the change deserves. Adding one field renumbers the positional variables in the
generated `sdk.go` (`f15` becomes `f16`, and so on), which is most of the line count and none of
the behaviour. The rest is moving the SDK model from v1.41.0 to v1.41.10, the earliest release
with the durability model, which refreshes doc comments and read-only status fields on
neighboring resources. 

Thanks @knottnt for the find-only approach in #230 — this adopts it.

By submitting this pull request, I confirm that my contribution is made under the terms of the
Apache 2.0 license.


test/e2e/tests/test_replicationgroup.py
```diff
@@ -1,4 +1,5 @@
 # Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+# Copyright © 2026 Apple Inc.
```
knottnt: I don't think we can share the copyright here.

config/crd/bases/elasticache.services.k8s.aws_serverlesscaches.yaml
```diff
@@ -112,6 +112,13 @@ spec:
                   The version of the cache engine that will be used to create the serverless
                   cache.
                 type: string
+              networkType:
```
knottnt: This appears to be an additional new spec field. Would you be able to include an e2e test validating the create/update behavior for this field as well for the serverless cache tests? Otherwise, we can defer adding this new field by adding it to the ignored field paths in generator.yaml.

config/crd/bases/elasticache.services.k8s.aws_snapshots.yaml
```diff
@@ -283,6 +285,12 @@ spec:
                   using the r6gd node type. This parameter must be set to true when using r6gd
                   nodes. For more information, see Data tiering (https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/data-tiering.html).
                 type: string
+              durability:
```
knottnt: This appears to be an additional new spec field. Would you be able to include an e2e test validating the create/update behavior for this field as well for the snapshot tests? Otherwise, we can defer adding this new field by adding it to the ignored field paths in generator.yaml.

pkg/resource/replication_group/hooks.go
```diff
@@ -1407,6 +1408,23 @@ func modifyDelta(
...
+	// We add it here so that explicit transitions (sync<->async) are detected.
+	// nil latest means the cluster predates the Durability feature (effectively
+	// disabled). Adding the delta lets the disabled guard in
+	// sdk_update_pre_build_request reject the attempt with a terminal error.
+	if ackcompare.HasNilDifference(desired.ko.Spec.Durability, latest.ko.Spec.Durability) {
+		delta.Add("Spec.Durability", desired.ko.Spec.Durability, latest.ko.Spec.Durability)
+	} else if desired.ko.Spec.Durability != nil && latest.ko.Spec.Durability != nil {
+		d := *desired.ko.Spec.Durability
+		l := *latest.ko.Spec.Durability
+		if d != l && d != "default" && l != "default" {
```
knottnt: Q: If the ReplicationGroup is created with a "default" durability value will latest not always have "default" as its value? If it does will the `l != "default"` check not always block drift detection here and prevent the user from updating to a non-default value?

templates/hooks/replication_group/sdk_update_pre_build_request.go.tpl
```diff
@@ -3,3 +3,18 @@
...
+		latestDurability := ""
+		if desired.ko.Spec.Durability != nil {
+			desiredDurability = *desired.ko.Spec.Durability
+		}
+		if latest.ko.Spec.Durability != nil {
+			latestDurability = *latest.ko.Spec.Durability
+		}
+		// Treat nil latest as "disabled" — clusters that predate the Durability
+		// feature are effectively disabled and cannot have durability enabled.
+		if desiredDurability == "disabled" || latestDurability == "disabled" || latestDurability == "" {
```
knottnt: Q: Does the AWS API not perform this validation itself? If it does what's the reason for not deferring validation to the service?

knottnt (commented):
Thanks @n-bezawada. Left a few comments.

test/e2e/tests/test_replicationgroup.py
n-bezawada: just this file or across the board ?

knottnt: I didn't see the Apple copyright in any of the files. So, I think this is the only place it needs to be removed. But generally this would apply to any repos under `aws-controller-k8s` GitHub org. 
```suggestion
```

n-bezawada: @knottnt  unfortunately our legal/oss teams required * this , checking with them to see what the next steps maybe If we can't find a path forward  -  I understand if this won't work for you  and I can close this PR and perhaps y'all can implement this .

config/crd/bases/elasticache.services.k8s.aws_snapshots.yaml
n-bezawada: added to the ignored field path

pkg/resource/replication_group/hooks.go
n-bezawada: similar lines as this https://github.com/aws-controllers-k8s/elasticache-controller/pull/223#discussion_r3709389333 , AWS APIs don't allow durability to be toggled from default ( false ) to something else once the resource has been created.

n-bezawada: if the convention is to solely rely on AWS APIs can remove this

config/crd/bases/elasticache.services.k8s.aws_serverlesscaches.yaml
n-bezawada: added to the ignored field path

test/e2e/tests/test_replicationgroup.py
n-bezawada: our team came back with - 

"You may add Your own copyright statement to Your modifications and may provide additional or different license terms and conditions for use, reproduction, or distribution of Your modifications, or for any such Derivative Works as a whole, provided Your use, reproduction, and distribution of the Work otherwise complies with the conditions stated in this License.” 

pet that - the copyright statement is allowed under the Apache 2.0 license governing the project.

https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/LICENSE#L124

jlbutler: Yeah this isn't really a licensing question, Apache 2.0 certainly allows this, and we don't have any technical issue with it.  It's more a project status quo and standard. 

The project has had contributions from dozens of other companies including Apple, and we've not added additional copyrights prior to this. Is this a hard requirement for your contribution?

pkg/resource/replication_group/hooks.go
knottnt: Generally, we defer to the AWS API when it comes to service level validation. Duplicating the service's validation risks drift between what the service actually allows and ACK. Additionally, in this case it looks like we're also suppressing the delta from being added. If I'm understanding this correctly that would result in the an illegal "default" -> "async/sync" update being silently ignored instead of throwing an error alerting the user that their change can't be applied to the AWS resource.

test/e2e/tests/test_replicationgroup.py
n-bezawada: @jlbutler  thanks for the review ,   its a hard requirement per the legal team.

jlbutler: Thanks for checking with your team, and we understand the ask.

Just to level-set for posterity: IANAL, but the notice isn't strictly necessary here. You automatically own your contribution the moment you submit it (we have no CLA), so a copyright line documents that ownership rather than creating it.

But if you'd like to publicly record Apple's authorship of this contribution, that's completely fine and we're happy to support it.

That said, I'd prefer to keep it out of individual source files. Landing copyright banners in source is a precedent I want to avoid for maintenance reasons (they proliferate and drift over time, if we refactor code in the future we need to reason about migration of notices, etc).

The Apache-native home for this is /NOTICE, and I think it's actually better for your ask. It announces the contribution across the whole repo, in one durable place, rather than on a single test file.

Could you move the addition there instead? Something like:

  Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.

  + Portions Copyright 2026 Apple Inc.

That's a precedent I'm glad to set. I'll cut a PR in our Contributor docs to formally document this, so other contributors can follow the same pattern.
  
If that works for you, please update your PR and we'll continue toward merging your contribution (thanks again for taking the time!).

n-bezawada: Our team has confirmed that this is acceptable , made this change.

pkg/resource/replication_group/hooks.go
n-bezawada: good call , removed  .

pkg/resource/replication_group/hooks.go
```diff
@@ -1407,6 +1407,31 @@ func modifyDelta(
...
+	// We therefore re-add ANY difference to the delta -- including transitions
+	// to/from "default" -- so ModifyReplicationGroup is actually called.
+	// Whether durability is enabled or disabled is fixed at cluster creation
+	// and AWS does not support changing that afterward; such requests are
+	// rejected by the service itself (InvalidParameterCombination or
+	// InvalidParameterValue, both already in terminal_codes below) rather than
+	// validated client-side here.
+	// nil latest means the cluster predates the Durability feature (effectively
+	// disabled).
+	if ackcompare.HasNilDifference(desired.ko.Spec.Durability, latest.ko.Spec.Durability) {
```
knottnt: nit: this looks like it matches the auto-generated delta comparison. If we no longer need custom comparison code for this field I think we can just remove the `compare: is_ignored:true` config in generator.yaml.

n-bezawada:
> @n-bezawada: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> elasticache-verify-code-gen	[eff0f2b](https://github.com/aws-controllers-k8s/elasticache-controller/commit/eff0f2b1ceb0b728e4f5e51a604f07cc84832414)	[link](https://prow.ack.aws.dev/view/s3/ack-test-infra-prod-prow-logs-453735116143/pr-logs/pull/aws-controllers-k8s_elasticache-controller/223/elasticache-verify-code-gen/2085813320652165120)	true	`/test elasticache-verify-code-gen`
> [Full PR test history](https://prow.ack.aws.dev/pr-history?org=aws-controllers-k8s&repo=elasticache-controller&pr=223). [Your PR dashboard](https://prow.ack.aws.dev/pr?query=is:pr+state:open+author:n-bezawada).
> 
> Details
> Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes-sigs/prow](https://github.com/kubernetes-sigs/prow/issues/new?title=Prow%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).

This test seems to fail coz of the copyright addition discussed [here](https://github.com/aws-controllers-k8s/elasticache-controller/pull/223#discussion_r3736461302)

knottnt (commented):
@n-bezawada Thanks for updates. Just had a minor nit regarding the custom delta code.

For the e2e test failures it looks like the problem is that boto3 isn't picking refreshed credentials during the extended test run. I believe updating [acktest](https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/test/e2e/requirements.txt) to `2efd6c7a98a1e99346f1de6c94f50f6bd5c329d8` will resolve that issue.

pkg/resource/replication_group/hooks.go
n-bezawada: done

n-bezawada:
> @n-bezawada Thanks for updates. Just had a minor nit regarding the custom delta code.
> 
> For the e2e test failures it looks like the problem is that boto3 isn't picking refreshed credentials during the extended test run. I believe updating [acktest](https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/test/e2e/requirements.txt) to `2efd6c7a98a1e99346f1de6c94f50f6bd5c329d8` will resolve that issue.

done.

knottnt:
@n-bezawada The update for the acktest library has resolved the ExpiredToken issue we were seeing, but it looks like some of the new e2e tests are failing their assertions. 

```
test_rg_durability_update:
             resource = k8s.get_resource(reference)
>           assert resource['spec']['durability'] == "sync"
E           AssertionError: assert 'async' == 'sync'
E             
E             - sync
E             + async
E             ? +

```
The above may be due to the ModifyReplicationGroup returning a stale value in its response. If you can confirm that is the reason for this failure we can omit setting Spec.Durability from the response with the below field config in generator.yaml.

```
set:
    - method: Update
       ignore: from
```



Both `test_rg_durability_default_guard` and `test_rg_durability_disabled_guard` are failing to observe ACK.Terminal. Would you be able to confirm what error code, if any, ModifyReplicationGroup. us returning when these two guards are violated?
```
assert k8s.wait_on_condition(
                reference, "ACK.Terminal", "True", wait_periods=30)
E               AssertionError: assert False
E                +  where False = <function wait_on_condition at 0x7f350595e840>(CustomResourceReference(group='elasticache.services.k8s.aws', version='v1alpha1', plural='replicationgroups', name='rg-durability-default-guard-3uh4', namespace='default'), 'ACK.Terminal', 'True', wait_periods=30)
E                +    where <function wait_on_condition at 0x7f350595e840> = k8s.wait_on_condition

```

knottnt:
For the `elasticache-verify-codegen` failure that is a non-issue. I've created a [PR](https://github.com/aws-controllers-k8s/code-generator/pull/729) to prevent our code-generator from overwriting the updated NOTICE file.

n-bezawada:
> @n-bezawada The update for the acktest library has resolved the ExpiredToken issue we were seeing, but it looks like some of the new e2e tests are failing their assertions.
> 
> ```
> test_rg_durability_update:
>              resource = k8s.get_resource(reference)
> >           assert resource['spec']['durability'] == "sync"
> E           AssertionError: assert 'async' == 'sync'
> E             
> E             - sync
> E             + async
> E             ? +
> ```
> 
> The above may be due to the ModifyReplicationGroup returning a stale value in its response. If you can confirm that is the reason for this failure we can omit setting Spec.Durability from the response with the below field config in generator.yaml.
> 
> ```
> set:
>     - method: Update
>        ignore: from
> ```

@knottnt  - 


Thanks for the pointers **On the stale response.** Confirmed against the live API, with the controller out of the picture:

```
$ aws elasticache modify-replication-group --replication-group-id <rg> --durability sync --apply-immediately
{
    "ReplicationGroup": {
        "Status": "available",
        "PendingModifiedValues": {},
        "Durability": "async",
        "EffectiveDurability": "async"
    }
}
```

Asked for `sync`, got `async` back. Same result 3 out of 3 times. `DescribeReplicationGroups` is briefly stale after the write as well — I saw +4s on one run and +17s on another, polling once a second.

**On scoping it to `Update`.** I tried that first and `test_rg_durability_update` failed the same way. `sdkFind` also populates `Spec.Durability`, and since the read is stale too it just puts the old value straight back. It needs to cover every operation:

```yaml
Durability:
  set:
    - ignore: from
```

**On the two guards.** With `Spec.Durability` being overwritten from the response, `desired == latest`, so no delta was computed, `sdkUpdate` was never called, and nothing ever reached AWS. There was no error to turn into `ACK.Terminal` — the modify simply never happened.

**The error codes you asked about.**

| Transition | Error |
|---|---|
| `default -> sync` / `default -> async` | `InvalidParameterCombination: Durability cannot currently be enabled on non-durable clusters` |
| `sync -> disabled` | `InvalidParameterCombination: Durability cannot currently be disabled on durable clusters` |

Both use `InvalidParameterCombination`   in `terminal_codes`, so the terminal condition takes care of itself once the request actually gets sent. No client-side validation went in — the e2e tests assert on the service's own message.


**How the comparison works now.** Against the durability the API reports, surfaced as `status.observedDurability`. The API reports it unresolved, so a group created with `default` reads back `default` — which is why `effectiveDurability` can't be used for this, its enum has no `default`. Same shape as `AccessString` on the `User` resource. I called it `observedDurability` since it holds what the API reports rather than what we last sent, but happy to rename it to match the `LastRequested*` naming if you'd prefer.

One side effect worth mentioning: because the comparison is against what the API reports instead of state we cache ourselves, a durability changed outside of ACK now gets corrected on the next reconcile. There's an e2e test for it

pkg/resource/replication_group/hooks.go
n-bezawada: Re-opening this , since 

`ModifyReplicationGroup` returns a stale `Durability` in its response, so populating `Spec.Durability` from it reverts whatever the user asked for.

n-bezawada: To close the loop on why the config has to come back: the fix for that stale   response is `set: ignore: from`, and it has to apply to every operation, not just   Update — `sdkFind` populates `Spec.Durability` too, and the read is briefly stale   after a write as well, so scoping it to Update alone still let the old value back in.    Once no response populates `Spec.Durability`, `latest.Spec.Durability` is never set,   so the generated comparison ends up diffing the desired value against `nil` on every   reconcile. That's why `compare: is_ignored: true` and the custom comparison are back —   they're a consequence of the fix rather than a preference. The comparison now runs   against `status.observedDurability`, the value the API reports.

pkg/resource/replication_group/sdk.go
```diff
@@ -157,11 +157,6 @@ func (rm *resourceManager) sdkFind(
 		} else {
 			ko.Spec.Description = nil
 		}
-		if elem.Durability != "" {
```
knottnt: I don't think we want to avoid setting Spec.Durability in sdkFind. Doing so will prevent adopted resources from setting the value.

generator.yaml
```diff
@@ -206,6 +206,16 @@ resources:
           is_ignored: true
       NetworkType:
         is_immutable: true
+      Durability:
+        set:
+          - ignore: from # the modify API returns a stale value; see modifyDelta
+        compare: # removes the spec field from automatic delta comparison
+          is_ignored: true
+      ObservedDurability:
```
knottnt: I'm not sure that `ObservedDurability` is necessary to avoid a stale value being written to the CR. While `sdkFind` does set `Spec.Durability` the object it returns represents "live AWS state" and is not used to write values back to the CR outside of resource adoption. Scoping `set:ignore` to only the Create/Update responses should be enough to avoid writing the stale value back.

pkg/resource/replication_group/hooks.go
```diff
@@ -622,14 +636,29 @@ func (rm *resourceManager) modifyReplicationGroup(
...
 		if desired.ko.Spec.Engine != nil {
 			normalizedRG.Engine = desired.ko.Spec.Engine
 		}
+		durabilityModified := delta.DifferentAt("Spec.Durability")
+		if durabilityModified {
+			if desired.ko.Spec.Durability != nil {
+				// Without this the durability just accepted by AWS would be reported as the
+				// previous value until the next read, both in status and in the difference
+				// computed on the following reconcile.
+				normalizedRG.Durability = svcsdktypes.Durability(*desired.ko.Spec.Durability)
```
knottnt: Not sure that this is necessary if we are already ignoring `Spec.Durability` in sdkUpdate with `set:ignore`. The value should already be the same as `desired` since the returned object is a deepcopy.

pkg/resource/replication_group/hooks.go
```diff
@@ -1188,6 +1217,13 @@ func (rm *resourceManager) newModifyReplicationGroupRequestPayload(
 		input.CacheParameterGroupName = desired.ko.Spec.CacheParameterGroupName
 	}
 
+	// the delta for this field is computed against the durability reported by the API in
+	// Status.ObservedDurability -- see durabilityRequiresUpdate
+	if delta.DifferentAt("Spec.Durability") &&
```
knottnt: Does durability need to be changed with these values special cased values? If not I think we can just use the generated `newUpdatePayload` to set this.

pkg/resource/replication_group/hooks.go
```diff
@@ -1395,6 +1431,13 @@ func modifyDelta(
 			unmarshalLastRequestedLDCs(desired))
 	}
 
+	// note that the comparison is done against latest.Status.ObservedDurability, the
+	// durability reported by the API, as opposed to latest.Spec.Durability, which is never set
+	// from the AWS response -- see the Durability field config in generator.yaml
+	if durabilityRequiresUpdate(desired, latest) {
```
knottnt: latest.Spec.Durability already represents the live value reported by the AWS Describe API. I don't we need to use a custom delta here to compare the desired spec value against a custom status field.

knottnt (commented):
> On scoping it to Update. I tried that first and test_rg_durability_update failed the same way. sdkFind also populates Spec.Durability, and since the read is stale too it just puts the old value straight back. It needs to cover every operation:

@n-bezawada I'm not sure that the root cause was that sdkFind populates `Spec.Durability`. While that is the case, the object returned by `sdkFind` represent observed state is not used to directly write values back to the desired manifest. I think the test failures you saw may have been due to `Status.EffectiveDurability` retaining the stale value since the ACK controller wasn't requeueing after the update. This behavior would've been fixed by the forced requeue you introduced by forcing the resource status to modifying. I've gone ahead and made a draft PR to demonstrate retaining `Spec.Durability` in `sdkFind` [here](https://github.com/aws-controllers-k8s/elasticache-controller/pull/230).

pkg/resource/replication_group/sdk.go
n-bezawada: The adoption case is a good catch I'd missed. the suppression is now scoped to the create and modify responses only  based on your https://github.com/aws-controllers-k8s/elasticache-controller/pull/230

generator.yaml
n-bezawada: Confirmed, and `ObservedDurability` is gone

pkg/resource/replication_group/hooks.go
n-bezawada: Agreed, removed. The normalization was only there because I was suppressing the value on every operation, so nothing was populating it. With the suppression scoped to the write responses it isn't needed.

pkg/resource/replication_group/hooks.go
n-bezawada: Changed . Durability goes through the generated payload now

n-bezawada:
Reworked this to the find-only approach from #230 — thanks @knottnt, that's a simpler design and it should resolve your comments.

knottnt (approved):
Thanks @n-bezawada!

/lgtm

